PR #145, opened 2015-01-05
tool to update Solr documents with availability score

drh-stanford:
Addresses #8


mejackreed:
Could we pop this into a rake task?
`/lib/tasks/...`

drh-stanford:
I think a rake task would require Capistrano/puppet because `solr.yml` only has the read endpoint for the Solr index. I'm not sure you can send an update request to that URL. @eefahy? Can we update from the load balancer or do we need to talk to a specific cluster node? I've always done the latter.

mejackreed:
I think you can run a task like this:

```
$ SOLR_URL=http://my-solr-server:PORT/solr/core rake earthworks:update_score

drh-stanford:
Hmmm, I was thinking that you wanted a rake task that actually updated the correct index when run by default. I can make it configurable as you describe.

PR #151, opened 2015-01-12
tweek solr config to address geoblacklight/geoblacklight#258

drh-stanford:
See geoblacklight/geoblacklight#258

### DIFF
--- a/config/solr_configs/solrconfig.xml
+++ b/config/solr_configs/solrconfig.xml
@@ -94,10 +94,30 @@
       <int name="indent">2</int>
       <str name="defType">edismax</str>
       <str name="echoParams">all</str>
+      <str name="mm">6&lt;-1 6&lt;90%</str>
+      <int name="qs">1</int>
+      <int name="ps">0</int>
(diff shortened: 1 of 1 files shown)


drh-stanford:
`kurma-earthworks-dev` has this config change for testing

mejackreed:
This seems like a great candidate to add in an integration test for testing relevance of what should come back.

The tests can live under `spec/integration/external-data` 

See https://github.com/sul-dlss/earthworks/blob/master/spec/integration/external-data/author_search_spec.rb for an example.

What do you think?

drh-stanford:
I prefer to wait a bit until we get a better sense of what the relevancy ranking should be after hearing some comments from users. For example, I dunno what should be the top 10 hits for "New Zealand" -- the ranking looks somewhat arbitrary to me, e.g., maybe it should be sorted by score and most recent data. Another example is "new zeland lakes" reverses the rank order of the "new zeland" query but "lakes new zealand" preserves the order.

mejackreed:
:+1: Sounds good to me!

Issue #81, opened 2014-12-05
Add use and copyright information to metadata show page


drh-stanford:
Examples: 

Restricted via GIS Restricted APO: http://purl.stanford.edu/vv853br8653
Restricted via Maps of Africa APO: http://purl.stanford.edu/jc604kx9499
Public via GIS Public Domain APO: http://purl.stanford.edu/cz128vq0535

drh-stanford:
Turns out we have 3 APOs in our current corpus, so we'll need to parse the APOs to get the text, or something cleaner...

mejackreed:
I think we need to have something in the Solr document that specifies which APO to use. Since "Public" / "Restricted" isn't enough.

mejackreed:
So one thing also to note here is I don't think we need to actually add the copyright. Maybe we can get clarification? But we do not show copyright in SearchWorks only Use and Reproduction. @jvine can you comment on that?

drh-stanford:
@kimdurante hey Kim... Tom wanted to add the Use and Reproduction language to the Earthworks metadata, but the examples I've found use the DC `rights` field for this. Is there another field that makes sense for the use and reproduction language as we're already using the `rights` field for "Public" vs. "Restricted"... if not, do you have any suggestions on how we might put both into the `rights` field?

See http://www.alastore.ala.org/nealschuman/companionwebsite/metadata-digital-collections/MDC_DC-MODS_Element_Comparison_Examples.pdf

drh-stanford:
This issue was born out of the demo meeting from Tom's comments. That's where I got both use and copyright. I might have transcribed it wrong so we should get some clarification.

kimdurante:
Hey...One way to handle this might be to use DC rights field for the use and reproduction statement and use  'dct:accessRights' for expressing public vs. restricted. If you need a unique rights field to govern access functionality in Earthworks that would probably be the best approach.
See: http://dublincore.org/documents/dcmi-terms/#terms-accessRights
'rights' is sometimes a more general permission statement,  'accessRights' applies specifically to policies. 

If you don't need a unique field to designate public/restricted and don't want to modify the geoBL  schema, we could probably concatenate the two fields from the access policy. 

Example:
"Restricted: These data are licensed by Stanford Libraries and are available to Stanford University affiliates only. Affiliates are limited to current faculty, staff and students. These data may not be reproduced or used for any purpose without permission. For more information please contact brannerlibrary@stanford.edu."

mejackreed:
[...]

PR #162, opened 2015-01-15
adds in use and reproduction statement to upper metadata

### LINKED ISSUE #158: Add license attribution data for Stanford objects

mejackreed:
Closes #158 

#### stanford-cz128vq0535
screen shot 2015-01-15 at 10 32 55 am

#### stanford-cg357zz0321
screen shot 2015-01-15 at 10 32 44 am

### DIFF
--- a/app/views/catalog/_upper_metadata.html.erb
+++ b/app/views/catalog/_upper_metadata.html.erb
@@ -1,0 +1,49 @@
+<% document ||= @document %>
+<div class="geoblacklight-view-panel">
+  
+  <dl class="dl-horizontal">
+    <% if document['dc_creator_sm'].present? %>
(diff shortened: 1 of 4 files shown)


drh-stanford:
This should also have the copyright information per #81.

Unfortunately, I don't think the static text for Public vs. Restricted will work :(. We have more than 2 APOs in our content. There's a 3rd APO for the Africa Maps. See http://purl.stanford.edu/jc604kx9499

I need to circle back with Kim to figure this out.

mejackreed:
Ok to complete this then, there will need to be metadata added to the solr documents. This can be in a field that doesn't necessarily have to be created in geoblacklight-schema (but it might make sense as others will probably run into this).

PR #189, opened 2015-02-18
update some heading on home page

mejackreed:
screen shot 2015-02-18 at 3 42 43 pm


mejackreed:
@drh-stanford good point. took your suggestion and used "here"

PR #192, opened 2015-02-20
adds in message if no downloads are available

### LINKED ISSUE #188: Show alert to tell user to click feedback if layer's data not downloadable
because of a lack of WFS/direct links

mejackreed:
Addresses concerns raised in ad hoc meeting about letting the user know that a layer is unavailable to download if no download options are present.

Deployed to dev successfully 

Closes #188

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -1,2 +1,8 @@
 module ApplicationHelper
+  def document_unavailable_options(document)
+    options = []
+    options.push 'download' unless document.direct_download || document.download_types.present?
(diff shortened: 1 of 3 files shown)


mejackreed:
My goal here was to have the message as generalizable as possible to account for other messages in the future. I also didn't want to cause confusion by introducing a new statement to a user:

"This data is hosted by Institution, and is currently unavailable to preview and download...."
"This data is hosted by Institution, and they do not offer direct download for this data... "

This could be potentially problematic for if we get something down the road that has a preview available but not a download.

PR #196, opened 2015-02-24
upgrade to Solr 4.10.3

drh-stanford:


drh-stanford:
I can reproduce this Travis error on my laptop. I will resubmit...

PR #202, opened 2015-02-26
tool that generates a sitemap.xml.gz based on slugs in Solr index

drh-stanford:
This tool generates a `sitemap.xml.gz` with monthly frequency for all the slugs in the Solr index. To use it run:

```
cd tools/sitemap
bundle exec ./generate.rb https://sul-solr.stanford.edu/solr/kurma-earthworks-prod http://earthworks.stanford.edu
```

then add it to your `robots.txt`:

```
Sitemap: public/sitemap.xml.gz
```

see https://support.google.com/webmasters/answer/183668?hl=en&ref_topic=6080646&rd=1 for details


mejackreed:
Is it possible to convert this to a rake task rather than a Ruby script? That way we are not maintaining separate Gemfiles and can run this on the prod boxs. 

Rake task should live under `lib/tasks`

mejackreed:
Also one benefit it allows us to easily automate this using the whenever gem. We could regenerate nightly... etc.

drh-stanford:
The sitemap generator integrates with Capistrano so I'll do that work. hold off on this pull request...

mejackreed:
:+1: .. one question... do we have to redeploy the app if the index changes to generate the sitemap? Or can we just kick off a rake task?

Looks awesome thanks!

drh-stanford:
they are all standalone rake tasks. it's optional to rebuild the indexes on deploy (i.e., you have to explicitly add 'deploy:sitemap:refresh' to your Cap deploy scriptie, which i'll leave out because the solr query to fetch all the slugs takes a while).

drh-stanford:
To use:

```
cd earthworks/current
bundle exec rake sitemap:create
vi public/robots.txt # add Sitemap: line
```

config/sitemap.rb
```diff
@@ -0,0 +1,36 @@
+#
```
mejackreed: Can this entire thing be moved to a Rake task? That will give us a lot of benefits like:

 - RSolr already available to us and configured for the specific environment
 - Use of rails_config gem for env specific configuration options

mejackreed: Something like this? https://github.com/sul-dlss/earthworks/pull/206

drh-stanford: If we want to use the rake tasks already provided by `sitemap_generator`, they're based off running the `config/sitemap.rb` file. You can change its location like so:

```
CONFIG_FILE="config/some_other_sitemap.rb" rake sitemap:refresh
```

but the code that builds the sitemaps is supposed to live in `config/sitemap.rb` by design.  See https://github.com/kjvarga/sitemap_generator/blob/master/lib/sitemap_generator/tasks.rb

We will only run the sitemaps in production so we shouldn't need any configuration outside of the `sitemap.rb` and we can put the `sitemap.rb` as a shared file rather than in github. 

Alternatively, I can implement a `sitemap_refresh` Rake task that copies their code into our own tasks. Thoughts?

mejackreed: Oh I gotcha... hm. I would ask then if we can configure it  like this?

https://github.com/sul-dlss/earthworks/pull/207

PR #203, opened 2015-02-27
modify format facet to actually be download as [WIP]

mejackreed:
Implements the described functionality here: https://github.com/geoblacklight/geoblacklight/issues/304/

I think this is much more usable than knowing what the original file format was. This also helps to eliminate potential bad data coming from other sources.

screen shot 2015-02-27 at 2 35 20 pm


mejackreed:
Deployed to dev

mejackreed:
Checking the solr response times, lets keep an eye on this

mejackreed:
Solr performance dev (with new facets):
Home page load: 436ms
Spatial search: 1544ms
Download as KMZ: 208ms
Spatial search to bangladesh: 2826.1ms

Solr performance dev (without new facets):
Home page load: 122ms
Spatial search: 123ms
format shapefile: 230ms
spatial search to bangladesh: 915ms

thatbudakguy:
we decided to just not support faceting on format at all, since there are now "download as" options in v4 by default and you can facet on the type of data you want more easily.

PR #211, opened 2015-03-04
Add saved search button to search results page

mejackreed:
Previously, it was quite difficult to actually find how to save searches. This is now easier as "Save this search" and "Forget this search" are in the results tools

screen shot 2015-03-04 at 3 16 51 pm
screen shot 2015-03-04 at 3 16 46 pm


drh-stanford:
When I search and get 0 results I still get the "Save this search" button. Is this the expected behavior?

mejackreed:
Yes I think that's still an appropriate behavior. Also force pushed with a rebase from master.

PR #275, opened 2015-07-09
more flexible syntax for license tags

drh-stanford:
allows `by_xx` to be used in addition to `by-xx`. fixes #247


mejackreed:
Question here... 

On http://purl.stanford.edu/bf137pp5814.xml `by_sa` is used

Here http://purl.stanford.edu/vv853br8653.xml `by-nc` is used

Is there a convention that we use (or should use)?

mejackreed:
I'm wondering, do we need this still?

PR #285, opened 2015-10-19
Log CartoDB usage information

mejackreed:


app/assets/javascripts/modules/analytics.js
```diff
@@ -75,4 +75,9 @@ Blacklight.onLoad(function() {
       window._gaq.push(['_trackEvent', 'Failed Download', data.downloadId, data.downloadType]);
     }
   });
+
+  // Log Open in CartoDB Clicks
+  $(document).on('click', 'li.exports a:contains("Open in CartoDB")', function(e) {
+    window._gaq.push(['_tractEvent', 'Open in CartoDB', window.location.pathname.replace('/catalog/', '')]);
```
drh-stanford: looks like there's a typo here: `_tractEvent` rather than `_trackEvent`?

mejackreed: Thanks for catching that. I've made the change and pushed the update.

mejackreed:
@drh-stanford should be good to go

PR #288, opened 2015-10-22
Create a Earthworks::SearchBuilder to make spatial search results mor…

mejackreed:
…e relevant

On some searches (test + spatial) global datasets that matched term search were consistently ranked higher and would bump out other results. I think this approach provides for a better result for user. Also added index integration test.


drh-stanford:
this looks fine but needs a merge first...

mejackreed:
Thanks... rebased.. ready to :ship:

PR #294, opened 2016-02-11
Update dependencies

mejackreed:


drh-stanford:
is this strictly an update to a later Blacklight release, or are you changing functionality?

mejackreed:
Strictly an updated except for the minor page title change using the "newer" way Blacklight names titles. Caught in specs: https://github.com/sul-dlss/earthworks/commit/e0895809a56a9895d47d26fc810ba649bcf84b8b

cbeer:
Is the "newer" way wrong or a bug upstream?

mejackreed:
Eh.. not a "bug". Just things changed. With no query submitted, the resultant index page title is kind of strange. But /shrug

Gemfile
```diff
@@ -60,8 +60,8 @@ gem 'pg'
 # gem 'debugger', group: [:development, :test]
 
 
-gem "blacklight", '~> 5.12.1'
-gem "geoblacklight", '~> 0.11.0'
+gem "blacklight"
+gem "geoblacklight", '~> 0.12.1'
```
drh-stanford: is there a reason we include blacklight here? doesn't geoblacklight already require blacklight?

mejackreed: Yep, we are also updating our blacklight version. Partly to make sure that GeoBlacklight in prod works for Blacklight 5.18 in anticipation of updating to 6.0x

PR #301, opened 2016-05-01
bundle without development for stage and prod

mejackreed:
Spring and other things are being deployed in prod. This fixes that which hopefully should reduce some of the memory footprint.


drh-stanford:
Hmmm not sure why we didn't catch that sooner 👍

mejackreed:
I know right :)

PR #306, opened 2016-07-06
Update basemap

mejackreed:
This PR updates the basemap to use the HOT OSM basemap. I also modified the hover and bounding color to use cardinal red. I felt like without that, there was just too much blue going on.

screen shot 2016-07-06 at 2 11 20 pm
screen shot 2016-07-06 at 2 11 03 pm


mejackreed:
ping @mapninja

PR #322, opened 2016-08-05
Migrate to solrwrapper & update to GeoBlacklight v1.1.0

mejackreed:


drh-stanford:
Is it possible to use the GeoBlacklight `schema.xml` as-is rather than are own  `config/solr_configs/schema.xml` for development/test? i.e., copy it over via rake?

Also, we'll need to upgrade our fixture data to the v1 schema at some point. That may mean we have to build a v1 Solr index for earthworks -- I prefer to wait for feedback from Branner on the -dev v1 index, but that timeline is uncertain.

mejackreed:
Closing PR in favor of a smaller scoped one #328

PR #358, opened 2017-12-11
fix deploy packages and set dev/stage to RAILS_ENV=production

drh-stanford:
@mejackreed this cleans up the deploys a bit and puts stage into production mode

### DIFF
--- a/config/deploy/dev.rb
+++ b/config/deploy/dev.rb
@@ -1,6 +1,6 @@
-set :bundle_without, %w[sqlite production test].join(' ')
+set :bundle_without, %w[test development deployment].join(' ')
 
 server 'kurma-earthworks-dev.stanford.edu', user: 'geostaff', roles: %w[web db app]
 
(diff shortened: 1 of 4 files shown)


config/deploy/dev.rb
```diff
@@ -1,4 +1,4 @@
-set :bundle_without, %w[sqlite production test].join(' ')
+set :bundle_without, %w[production].join(' ')
```
mejackreed: Does this mean `pg` is not available on dev?

drh-stanford: yes. we should probably switch -dev to use RAILS_ENV=production too. I only use dev to test the index. what are you using -dev for, if anything?

mejackreed: Just to randomly test application code. I'm 👍 for the change.

drh-stanford:
ok, i've switched both dev and stage to use RAILS_ENV=production

PR #370, opened 2017-12-19
GeoMonitor integration directly into EarthWorks.

mejackreed:


db/migrate/20171215154810_create_geo_monitor_statuses.geo_monitor.rb
```diff
@@ -0,0 +1,13 @@
+# This migration comes from geo_monitor (originally 20171212181423)
+class CreateGeoMonitorStatuses < ActiveRecord::Migration[5.1]
+  def change
+    create_table :geo_monitor_statuses do |t|
+      t.string :res_code, index: true
```
drh-stanford: why do we need a single-field index on the `res_code`?

db/migrate/20171215154809_create_geo_monitor_layers.geo_monitor.rb
```diff
@@ -0,0 +1,16 @@
+# This migration comes from geo_monitor (originally 20171212180244)
+class CreateGeoMonitorLayers < ActiveRecord::Migration[5.1]
```
drh-stanford: why do we need single-field indexes on `access` and `checktype`?

lib/tasks/earthworks.rake
```diff
@@ -59,4 +59,25 @@ namespace :earthworks do
       File.open(fn, 'w') {|f| f << JSON.pretty_generate(data) }
     end
   end
+  namespace :geomonitor do
+    desc 'Update the layers GeoMonitor checks from the Solr index'
+    task update: [:environment] do
+      response = Blacklight.default_index.connection.get(
+        'select',
+        params: { q: '*:*', fl: '*', rows: 999_999 }
```
drh-stanford: Does Blacklight/RSolr provide a paging API that we could use instead to downloading all data in one query?

config/environments/production.rb
```diff
@@ -58,7 +58,7 @@
   # config.cache_store = :mem_cache_store
 
   # Use a real queuing backend for Active Job (and separate queues per environment)
-  # config.active_job.queue_adapter     = :resque
+  config.active_job.queue_adapter     = :sidekiq
```
drh-stanford: We should probably configure the number of sidekiq workers in our deploy/prod.rb and start small

app/models/concerns/status_extension.rb
```diff
@@ -0,0 +1,18 @@
...
+    after_commit :update_index
+  end
+
+  def update_index
+    data = [{
+      layer_availability_score_f: { set: layer.availability_score },
+      layer_slug_s: layer.slug
+    }]
+    Blacklight.default_index.connection.update(
+      params: { commitWithin: 500, overwrite: true },
```
drh-stanford: I would either choose a larger commitWithin (say 5 seconds) or make this configurable. I don't think we need a fast commit for the availability_score, right?

mejackreed: Nope not at all. I'll make this configurable with a larger default

db/migrate/20171215154809_create_geo_monitor_layers.geo_monitor.rb
mejackreed: So this was coming upstream from GeoMonitor. At the moment I'm not sure we do need these, but it may be helpful (perhaps?) if we need to query by them at some point. Would you prefer we remove them to begin with?

lib/tasks/earthworks.rake
mejackreed: Great suggestion, I'll look into it.

db/migrate/20171215154809_create_geo_monitor_layers.geo_monitor.rb
drh-stanford: yeah i would take them out and tune for performance later. usually you need composite keys for status-like fields for them to be used efficiently -- i.e., `where(access: 'Public', checktype: 'whatever')` benefits more from a composite key on access + checktype.

mejackreed: 👍

app/jobs/check_layer_job.rb
```diff
@@ -4,6 +4,15 @@ class CheckLayerJob < ApplicationJob
   ##
   # @param [GeoMonitor::Layer] layer
   def perform(layer)
-    layer.check
+    # See if in Solr first, if not, deactivate and exit
+    num_found = Blacklight.default_index.connection.get(
+      'select',
+      params: { q: "layer_slug_s:#{layer.slug}", fl: '*' }
+    )['response']['numFound'].to_i
+    if num_found > 0
```
mejackreed: @drh-stanford I would like your advice on this approach. This seemed like a better approach than trying to query the RDBMS with values in Solr (a very large SQL select that I think was bogging down the previous implementation).

drh-stanford: The Solr index is the authoritative place to look, so I think this is fine to query by the slug during the job. Although you may not need the `fl: '*'` -- how about returning only the slug itself, unless you need additional information.

PR #402, opened 2018-01-24
Homepage update

mejackreed:
This PR does the following things:

 - Adds byebug for debugging
 - Updates the homepage for improved visual design
   - Features the search bar more prominently with a new call to action
   - Removes all facet links in favor of important featured searches
   - Adds more context to map, the aim here is to help a non-technical user

screen shot 2018-01-24 at 9 25 58 am


app/models/concerns/featured_content_behavior.rb
```diff
@@ -0,0 +1,41 @@
...
+    end
+
+    solr_params
+  rescue NoMethodError
+    solr_params
+  end
+
+  private
+
+  def scanned_maps_content_params
```
drh-stanford: what do the Blacklight breadcrumbs look like for these queries?

config/locales/blacklight.en.yml
```diff
@@ -5,10 +5,10 @@ en:
       form:
         search:
           label: 'search for'
-          placeholder: ''
+          placeholder: 'Try "groundwater"'
```
drh-stanford: I find this helper tip a little confusing -- are there other suggestions?

app/models/concerns/featured_content_behavior.rb
mejackreed: screen shot 2018-01-24 at 11 47 18 am

config/locales/blacklight.en.yml
mejackreed: perhaps just "groundwater" or leave it blank entirely?

drh-stanford: how about "Search for data and maps"? or blank is ok with me. The former begs the question whether we need that header. This isn't my forte though :).

mejackreed: Yep I just stole this from Airbnb:

screen shot 2018-01-24 at 12 25 05 pm

I'm inclined to leave it as is.. and see if we get any feedback. It's a quick change if we have to modify it.

PR #404, opened 2018-01-24
setup geomonitor to run periodically using whenever gem

### LINKED ISSUE #401: Add "cron job" or something for running GeoMonitor periodically
We want to run the `update` and `check_stanford`, `check_public` tasks at some regular interval. We can use the `whenever` gem to do this I think. We just need to define the frequency (i.e., semi-daily, daily, weekly, etc.)

mejackreed:
Closes #401 

I haven't tested this yet, but I used the roles so that we weren't enqueuing jobs from each node.

### DIFF
--- a/config/schedule.rb
+++ b/config/schedule.rb
@@ -1,0 +1,34 @@
+# Use this file to easily define all of your cron jobs.
+#
+# It's helpful, but not entirely necessary to understand cron before proceeding.
+# http://en.wikipedia.org/wiki/Cron
+
+# Example:
+#
+# set :output, "/path/to/my/cron_log.log"
+#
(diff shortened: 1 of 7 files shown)


config/schedule.rb
```diff
@@ -0,0 +1,34 @@
...
+
+every 1.day, roles: %i[whenevs] do
+  rake 'earthworks:geomonitor:update'
+end
+
+every 6.hours, roles: %i[whenevs] do
+  rake 'earthworks:geomonitor:check_stanford'
+end
+
+every 6.hours, roles: %i[whenevs] do
```
drh-stanford: this should be 48.hours i think...

PR #406, opened 2018-01-25
modify the unavailability messaging

### LINKED ISSUE #333: Users are ignoring licensing and distribution information
Users are providing feedback to try and obtain licensed data from EarthWorks. This happens for both Stanford and other institution data. What can we do to make this more obvious to users.

### LINKED ISSUE #346: For unavailable items, is it possible to link to provider email address in information?

mejackreed:
Closes #333 
Closes #346 

screen shot 2018-01-25 at 12 38 23 pm

### DIFF
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -96,3 +96,59 @@ LEAFLET:
       <<: *opacity_control
     IMAGEMAPLAYER:
       <<: *opacity_control
+
+INSTITUTIONS:
+  Columbia:
+    department: Columbia University GIS & Statistical Data Resources
+    email: dssc@library.columbia.edu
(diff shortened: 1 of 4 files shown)


config/settings.yml
```diff
@@ -96,3 +96,59 @@ LEAFLET:
       <<: *opacity_control
     IMAGEMAPLAYER:
       <<: *opacity_control
+
+INSTITUTIONS:
```
drh-stanford: are we missing `Stanford` here?

mejackreed: Nope.. I don't think it will ever get called based on our logic that will render this.

PR #461, opened 2018-08-08
space out geomonitor and reduce tolerance

mejackreed:
This hopes to alleviate some production problems with our checking of our geoserver instances


lib/tasks/earthworks.rake
```diff
@@ -84,21 +84,21 @@ namespace :earthworks do
     task check: [:environment] do
       GeoMonitor::Layer.where(active: true).find_each(batch_size: 250) do |layer|
         puts "Enqueueing check for #{layer.slug}"
-        CheckLayerJob.set(wait: [*1..25].sample.seconds).perform_later(layer)
+        CheckLayerJob.set(wait: [*60..1800].sample.seconds).perform_later(layer)
```
jkeck: This seems fine to me in general.  I'm kind of curious of the existing behavior though.  What is the rationale of randomizing the wait times for these jobs?

PR #494, opened 2019-01-07
Repin bootstrap until OS situtation is sorted

camillevilla:
😢


cbeer:
The "OS situation" is https://github.com/sul-dlss/operations-tasks/issues/1824.

PR #521, opened 2019-05-21
Unpin some gems and update

jkeck:


Gemfile
```diff
@@ -10,8 +10,6 @@ gem 'sass-rails', '~> 5.0'
 gem 'uglifier', '>= 1.3.0'
 # Use CoffeeScript for .js.coffee assets and views
 gem 'coffee-rails', '~> 4.2'
-# See https://github.com/sstephenson/execjs#readme for more supported runtimes
-gem 'therubyracer',  platforms: :ruby
```
jkeck: This seemed like it should be okay to do as I _think_ we have node on our rails apps now, but I can kill this if that's not the case or we're using `therubyracer` for something else in the app

mejackreed: Yes should be fine. I've already removed in #514

jkeck: Okay, I ended up killing this commit in this PR because I don't know for sure if there was a js runtime on the earthworks servers (and didn't want to break the deploy accidentally).

PR #524, opened 2019-05-28
Add a limit to strings that should be calculated for suggestions

mejackreed:
Fixes an issue where autocomplete stopped working. Related to https://github.com/projectblacklight/blacklight/issues/1972


cbeer:
👍, and you'll make corresponding PRs to sul-solr-configs?

mejackreed:
Yep! https://github.com/sul-dlss/sul-solr-configs/pull/192

PR #565, opened 2019-10-14
Enable uglifier harmony mode

cbeer:


config/environments/production.rb
```diff
@@ -23,7 +23,7 @@
   config.public_file_server.enabled = ENV['RAILS_SERVE_STATIC_FILES'].present?
 
   # Compress JavaScripts and CSS.
-  config.assets.js_compressor = :uglifier
+  config.assets.js_compressor = Uglifier.new(harmony: true)
```
jkeck: Harmony mode is indicated as "experimental" by Uglifier.  I've poked around w/ it locally in some projects and haven't run into any issues, but are we sure this is production ready?

cbeer: 🤷‍♂ i think that ship sailed when Blacklight started using ES6.

PR #574, opened 2019-11-27
Script used to remediate rasters that have corrupted styles

mejackreed:
Used to resolve the data corruption documented in https://github.com/sul-dlss/gis-robot-suite/issues/224


ndushay:
Thank You!

PR #597, opened 2020-03-06
Add rack-attack to throttle requests

mejackreed:
Same implementation as SearchWorks.


jkeck:
You might be interested in implementing something like sul-dlss/SearchWorks#2374

mejackreed:
Thanks for that pointer. Just added that in.

PR #664, opened 2020-06-22
Check DOWNLOAD_PATH (when set) in OK Computer

### LINKED ISSUE #320: Settings.DOWNLOAD_PATH is configured but not used by is_it_working
In production, `Settings.DOWNLOAD_PATH` is configured but not used by is_it_working.

See https://github.com/sul-dlss/earthworks/blob/master/config/initializers/is_it_working.rb#L6

jkeck:
Closes #320

### DIFF
--- a/config/initializers/okcomputer.rb
+++ b/config/initializers/okcomputer.rb
@@ -2,4 +2,4 @@
 OkComputer.mount_at = false
 
 OkComputer::Registry.register "solr", OkComputer::HttpCheck.new(Blacklight.default_index.connection.uri.to_s.sub(/\/$/, '') + "/admin/ping")
-OkComputer::Registry.register "downloads-cache", OkComputer::DirectoryCheck.new((Rails.root + "tmp/cache/downloads").to_s, true)
+OkComputer::Registry.register "downloads-cache", OkComputer::DirectoryCheck.new(Settings.DOWNLOAD_PATH || (Rails.root + "tmp/cache/downloads").to_s, true)
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -61,6 +61,9 @@ RESTRICTED_URL: 'http://www.example-services.com'
 
 SOLR_COMMIT_WITHIN: 5000
 
+# DOWNLOAD_PATH is set in production but defaults to this tmp dir when not set
+# DOWNLOAD_PATH: './tmp/cache/downloads'
+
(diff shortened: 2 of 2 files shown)


config/initializers/okcomputer.rb
```diff
@@ -2,4 +2,10 @@
 OkComputer.mount_at = false
 
 OkComputer::Registry.register "solr", OkComputer::HttpCheck.new(Blacklight.default_index.connection.uri.to_s.sub(/\/$/, '') + "/admin/ping")
-OkComputer::Registry.register "downloads-cache", OkComputer::DirectoryCheck.new((Rails.root + "tmp/cache/downloads").to_s, true)
+
+if Settings.DOWNLOAD_PATH
+  OkComputer::Registry.register "downloads-cache", OkComputer::DirectoryCheck.new(Settings.DOWNLOAD_PATH, true)
```
jkeck, line 7: I'm a _weee_ bit nervous about this since we don't really have a way to test this in stage AFAICT.

mejackreed: Could we edit the config on the server to test?

jkeck: Sure, would you suggest temporarily setting a directory on stage that is diff from the temp dir?

Prod's dir is in `/var/cache`, should we just use the same dir on stage (permanently or temporarily)?

mejackreed: I'm not sure we want to set it to the same as prod.. just maybe even `/tmp/foo` and create that ?

jkeck: Okay, that seems to have worked on stage when I deployed this branch, configed a DOWNLOAD_PATH in tmp, and added the directory.

config/initializers/okcomputer.rb
```diff
@@ -2,4 +2,10 @@
...
+else
+  # What geoblacklight uses when no DOWNLOAD_PATH is set
+  OkComputer::Registry.register "downloads-cache", OkComputer::DirectoryCheck.new((Rails.root + "tmp/cache/downloads").to_s, true)
+end
```
cbeer, line 11: 🤷‍♂️ 

```suggestion
  OkComputer::Registry.register "downloads-cache", OkComputer::DirectoryCheck.new(Settings.DOWNLOAD_PATH || (Rails.root + "tmp/cache/downloads").to_s, true)
```

PR #668, opened 2020-06-23
Add a rake task to prune guest users and schedule to run nightly

### LINKED ISSUE #631: We need a prune user task that runs automatically
EarthWorks needs a prune user task to clean up guest users.

An alternative would be something similar to what SW does where guest users do not get created for bots.

jkeck:
Closes #631

### DIFF
--- a/config/schedule.rb
+++ b/config/schedule.rb
@@ -37,6 +37,10 @@ every '0 3 * * *', roles: %i[app] do # daily at 3 am
   rake 'earthworks:clear_rack_attack_cache'
 end
 
+every 1.day, at: '3:04 am' do
+  rake 'rake earthworks:prune_old_guest_user_data[3]'
+end
+
 every 1.day, at: '4:04 am' do
   rake 'rake earthworks:prune_old_search_data[14]'
 end
--- a/lib/tasks/earthworks.rake
+++ b/lib/tasks/earthworks.rake
@@ -146,6 +146,15 @@ namespace :earthworks do
                     'earthworks:opengeometadata:index']
   end
 
+  desc "Prune old guest users from the database"
+  task :prune_old_guest_user_data, [:months_old] => [:environment] do |t, args|
+    old_bookmarkless_guest_users_ids = User.guests_without_bookmarks
+                                           .where("users.updated_at < :date", { date: args[:months_old].to_i.months.ago })
+                                           .pluck(:id)
+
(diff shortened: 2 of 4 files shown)


lib/tasks/earthworks.rake
```diff
@@ -146,6 +146,15 @@ namespace :earthworks do
                     'earthworks:opengeometadata:index']
   end
 
+  desc "Prune old guest users from the database"
+  task :prune_old_guest_user_data, [:months_old] => [:environment] do |t, args|
+    old_bookmarkless_guest_users_ids = User.guests_without_bookmarks
+                                           .where("users.updated_at < :date", { date: months_old.to_i.months.ago })
```
jkeck, line 152: I suppose this could go into a scope as well, I waffled a bit.  Happy to do something else here.

mejackreed, line 152:
```suggestion
                                           .where("users.updated_at < :date", { date: args[:months_old].to_i.months.ago })
```

config/schedule.rb
```diff
@@ -37,6 +37,10 @@
   rake 'earthworks:clear_rack_attack_cache'
 end
 
+every 1.day, at: '3:04 am' do
+  rake 'rake earthworks:prune_old_guest_user_data[12]'
```
mejackreed, line 41: I'm wondering if even 3 months would be ok? maybe even less?

We are getting around 1.5 million every 3 months :(

lib/tasks/earthworks.rake
jkeck: 👍 I added this in the commit itself.

config/schedule.rb
jkeck: Yeah, totally.  I think, as mentioned in the issue, it might be a good idea to do something w/ a crawler detector to reduce the number of guest sessions, but I think that can be handled separately.

One issue I guess that does raise, is that this isn't going to be great at killing millions of database rows at once (and will be best to keep things manageable once they are at a manageable level).  This might benefit from operating on a table that is somewhat w/i a reasonable size. 

I'm not sure where it's gotten today, but maybe we should take a look at running the manual user pruning tasks we ran before to get things down before putting this in place?

mejackreed: I think thats a good idea. FWIW, I ran this on prod early morning and with 12 months it wasn't outrageously slow and completed withing 10 seconds.

PR #682, opened 2020-06-29
Remove blank field values when documents are indexed.

### LINKED ISSUE #644: Don't allow facets with empty values
These are reported as violations of "Ensure link text is meaningful within context". In each case, it's an empty facet value that creates an empty but perceivable (and operable!) link for a screenreader, and a broken display otherwise. We should prevent empty facet values.

https://earthworks.stanford.edu/catalog/facet/dc_publisher_s?featured=census_data&q=
https://earthworks.stanford.edu/catalog/facet/dc_publisher_s?featured=geospatial_data&q=
https://earthworks.stanford.edu/catalog/facet/dc_format_s?featured=geospatial_data&q=
https://earthworks.stanford.edu/catalog/facet/dc_publisher_s?f[dct_provenance_s][]=Harvard&featured=california_data&q=

jkeck:
Closes #644 

Adds an earthworks-processor that allows us to combine multiple processors.

Adds RemoveBlankFieldUpdateProcessorFactory to the earthworks-processor (in addition to the hashed id processor)

Note: these screenshots are using local fixtures and not production data.  I tested this w/ the ~4k NYU records locally and this fixed the issue for the 56 records that has blank publishers.

## columbia-columbia-landinfo-global-aet (before)
columbia-columbia-landinfo-global-aet-before

## columbia-columbia-landinfo-global-aet (after)
columbia-columbia-landinfo-global-aet-after

### DIFF
--- a/config/solr_configs/solrconfig.xml
+++ b/config/solr_configs/solrconfig.xml
@@ -38,22 +38,24 @@
     </autoCommit>
   </updateHandler>
 
-  <updateRequestProcessorChain name="add_hashed_id">
-    <processor class="solr.processor.SignatureUpdateProcessorFactory">
-      <bool name="enabled">true</bool>
-      <str name="signatureField">hashed_id_ssi</str>
-      <bool name="overwriteDupes">false</bool>
-      <str name="fields">layer_slug_s</str>
-      <str name="signatureClass">solr.processor.Lookup3Signature</str>
-    </processor>
+  <updateProcessor class="solr.processor.SignatureUpdateProcessorFactory" name="add_hashed_id">
+    <bool name="enabled">true</bool>
+    <str name="signatureField">hashed_id_ssi</str>
+    <bool name="overwriteDupes">false</bool>
+    <str name="fields">layer_slug_s</str>
+    <str name="signatureClass">solr.processor.Lookup3Signature</str>
+  </updateProcessor>
 
+  <updateProcessor class="solr.RemoveBlankFieldUpdateProcessorFactory" name="remove_blanks" />
+
+  <updateRequestProcessorChain name="earthworks-processor" processor="remove_blanks,add_hashed_id">
     <processor class="solr.LogUpdateProcessorFactory" />
     <processor class="solr.RunUpdateProcessorFactory" />
   </updateRequestProcessorChain>
 
   <requestHandler name="/update" class="solr.UpdateRequestHandler">
     <lst name="defaults">
-      <str name="update.chain">add_hashed_id</str>
+      <str name="update.chain">earthworks-processor</str>
     </lst>
   </requestHandler>
 
--- a/config/solr_configs/solrconfig.xml
+++ b/config/solr_configs/solrconfig.xml
@@ -38,17 +38,9 @@
     </autoCommit>
   </updateHandler>
 
-  <updateProcessor class="solr.processor.SignatureUpdateProcessorFactory" name="add_hashed_id">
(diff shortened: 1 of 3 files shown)


jkeck:
If this looks 👌 I'll get a PR for `sul-solr-configs`.  This will require that we re-run indexing of all the OGM data (as it only affects documents as they are indexed).

config/solr_configs/solrconfig.xml
```diff
@@ -38,22 +38,24 @@
...
 
-  <updateRequestProcessorChain name="add_hashed_id">
-    <processor class="solr.processor.SignatureUpdateProcessorFactory">
-      <bool name="enabled">true</bool>
-      <str name="signatureField">hashed_id_ssi</str>
-      <bool name="overwriteDupes">false</bool>
-      <str name="fields">layer_slug_s</str>
-      <str name="signatureClass">solr.processor.Lookup3Signature</str>
-    </processor>
+  <updateProcessor class="solr.processor.SignatureUpdateProcessorFactory" name="add_hashed_id">
```
mejackreed, line 41: I'm good with this approach, but I think we ended up removing the processor factory in prod.. and now I can't remember the reason why we did this. @cbeer do you remember?

cbeer: Yeah, it doesn't play nice with atomic updates.

mejackreed: Ah that was it. :(

jkeck: Ah, okay, I was wondering about that.  Does that mean that we can't use any processor factory (e.g. `RemoveBlankFieldUpdateProcessorFactory`?)

cbeer: I suspect it was specific to the `SignatureUpdateProcessorFactory`, but maybe filtering at index time would be better? (or, maybe without an XSLT in the mix, it's no longer an issue?)

jkeck: The content from the other institutions will still be an issue.  We index those records through a rake task provided by GeoBlacklight.  I'm not sure if we can dictate that all consumers of that indexing rake task should not receive the blank values or not, but 🤷‍♂️

mejackreed: This solution would handle other organizations metadata too, so if this worked w/ atomic updates I would be inclined to adopt something like this. Unless if when we look at reindexing other orgs, we can remove their empties then?

cbeer: 🤷‍♂️ only one way to find out.

jkeck: I'll deploy this branch to stage and re-index the content from external providers and see what happens. 🤞

cbeer: And make a PR to sul-solr-configs?

jkeck: definitely

jkeck: Re-ran indexing of OGM content in earthworks-stage and things appear to have worked well (assuming the issue isn't that atomic updates come around and blow things away later because of how I've configured the update handler).

## [nyu-2451-43780 (prod)](https://earthworks.stanford.edu/catalog/nyu-2451-43780)
nyu-2451-43780-before

## [nyu-2451-43780 (stage)](https://kurma-earthworks-stage.stanford.edu/catalog/nyu-2451-43780)
nyu-2451-43780-after

PR #684, opened 2020-06-29
Use maintained fork of twitter typeahead

### LINKED ISSUE #623: Fix suggestions
The autocomplete functionality no longer works. I _think_ this has something to do with analyzing a field that might be too large.

History: https://github.com/geoblacklight/geoblacklight/issues/422

 https://github.com/projectblacklight/blacklight/issues/1972

mejackreed:
Fixes #623 

Sigh.. so this has us using the "maintained" (really unmaintained) fork of the unmaintained project. While it does now finally work it is suboptimal overall.

We also are overriding the Blacklight search form partial to provide `autocomplete="off"` to the input element. The fork removed this for some unknown reason. Discussion upstream here: https://github.com/corejavascript/typeahead.js/issues/131

We could contribute the autocomplete="off" thing upstream to Blacklight? 🤷‍♂️


jkeck:
👍 to pushing upstream.  Seems like it should be relatively harmless for the autocomplete implementation we have and will be useful if we decide to use the "(un-)maintained" fork for any reason.  The only thing that gives me pause is if folks are using the built-in autocomplete functionality to inject their own customization (but I'm not sure we can/need to accommodate that case).

PR #692, opened 2020-07-02
Make the application name the h1; fixes #690

### LINKED ISSUE #690: Change <h1> construction to use visible banner title
There's an oddity with the `<h1>` that seems to be a Blacklight thing - it's also present in SearchWorks.

1. The visible application title (and home link) in the banner is not actually a heading.
2. There is a hidden `sr-only` `<h1>` further down in the DOM, which in some cases ends up in a bad location.

### Home page with the hidden `<h1>` exposed
Screen Shot 2020-07-01 at 5 35 09 PM

### Show page with the hidden `<h1>` exposed
Screen Shot 2020-07-01 at 5 37 32 PM

### Search results with several hidden headings exposed
This one is really broken - resulting structure is:
- `<h1>` Search
   - `<h2>` Search Constraints
- `<h1>` EarthWorks Geospatial Catalog
   - `<h2>` Search Results

Screen Shot 2020-07-01 at 5 43 51 PM

# Solution

- [ ] Make the visible title ALSO the `<h1>`, so that it is physically and logically at the top of the page. 

   If the title needs additional text for screen reader (e.g. "Geospatial Catalog") that can be included in a `<span>` within the heading. 

   This will
   - avoid drift between visible and hidden text, 
   - ensure that all pages get the heading in the same place, 
   - prevent the invisible heading from sliding down in the DOM as new features are added.

- [ ] On the search results page, remove the `<h1 class="sr-only top-content-title">Search</h1>`. It's not necessary and it's better to avoid more than one `<h1>` on a page.

cbeer:


mejackreed:
I'm seeing some spacing changes, is this just not syncing up right?

Screen Shot 2020-07-06 at 1 41 50 PM

cbeer:
Thanks; it looks like the sr-only text inside the link doesn't play nice with the text hiding?

Issue #619, opened 2020-05-05
Migrate MIT data to using new records


mejackreed:
Current EarthWorks MIT metadata was created from OGP records and given unique identifiers that are bespoke. We should migrate these records to using MIT's instance records at https://geodata.mit.edu/?f%5Bdct_provenance_s%5D%5B%5D=MIT&search_field=all_fields

It would be great to have redirects from old records to new ones, but I'm not sure its necessary. We should also verify that bookmarks are not broken.

jkeck:
I'm at least looking into what I can today (before vacation) and will report any details I find.

jkeck:
Here are some notes/proposals for what I found

* Records currently in EW via OGP: **7166**
* Records deposited in OGM: **1967** (**1965** successfully indexed locally [2 has bogus envelope data])
  * Able to map 1917 of the 7166 EW OGP records to these.
  * OGM then appears to provide us ~50 new records
* MIT Records in MIT GeoPortal: **10005**
  * Able to map 6921 of the 7166 EW OGP records to the records including "Mixed".
  * MIT's GeoPortal then appears to provide us ~3084 new records.
* We can use data in the layer_id_s/dct_references_s to match content that we had indexed before

## Proposal 1
* Index (1965) MIT records from OGM into EW
  * We can accomplish this simply by adding "edu.mit" to our local rake task
* Add code to handle redirect/whatever for the 1917 matching records
* Delete the 1917 matching records that was indexed from OGP
  * This would leave us with 5249 MIT records hanging around from OGP

## Proposal 2
* Index (10005) MIT records from the MIT GeoPortal into EW
  * We would need to write and maintain new code to do this, but might be useful for other content.
* Add code to handle redirect/whatever for the 6921 matching records
* Delete the 6921 matching records that we indexed from the MIT GeoPortal
  * This would leave us with 245 MIT records hanging around from OGP

A note on the code that would be required to redirect.  I experimented w/ this a bit a bit and this seemed to be able to be handled pretty straight forward with a class that loaded an ID map in a YAML file and can be used to redirect an old document to the new one in whatever way we like (I used a rescue_from block, but I'm sure there are other alternatives).  I have some code locally that can generate that ID map for us (from current `layer_slug_s` in the EarthWorks to the `layer_slug_s` that is in either the OGM data or MIT's GeoPortal).  I would be happy to get that pushed up into a PR if we think that either of these approaches are tractable.

jkeck:
[...]

PR #718, opened 2020-07-11
Add a rake task and configs to harvest MIT content from their …

jkeck:
…GeoBlacklight instance.

Part of #619

We can ship this once OpenGeoMetadata/GeoCombine#94 is merged

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,18 +1,4 @@
 GIT
-  remote: https://github.com/OpenGeoMetadata/GeoCombine.git
-  revision: 77fde5526c0d1c4d9b6c79d58cd0acb7351ce46b
-  branch: master
-  specs:
-    geo_combine (0.4.0)
-      activesupport
-      json-schema
-      net-http-persistent (~> 2.0)
-      nokogiri
-      rsolr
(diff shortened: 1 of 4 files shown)


Gemfile.lock
```diff
@@ -1,6 +1,6 @@
 GIT
   remote: https://github.com/OpenGeoMetadata/GeoCombine.git
-  revision: 77fde5526c0d1c4d9b6c79d58cd0acb7351ce46b
+  revision: b2688d4a75696e2211ae38f98989fe1a1c06542c
```
mejackreed, line 3: GeoCombine v0.5.0 has just been :shipit:

jkeck: Okay, do you want us to get on a released version?

mejackreed: Yeah we might as well

jkeck:
FWIW, I deployed this to stage and ran it (while temp. disabling SSL verification).  We have the [10,005 MIT records](https://kurma-earthworks-stage.stanford.edu/?f%5Bdct_provenance_s%5D%5B%5D=MIT) in stage.

PR #723, opened 2020-07-14
Update link hover styles to use a bottom border instead of …

### LINKED ISSUE #693: Contrast and identity colour changes for links, buttons, text
### 1. Link colour is too similar to the body text colour. 
There's no visual distinction of links until they are hovered; and then the hover state is slightly different from our other assets. (Instead of underline, we use border-bottom for better readability.)

- [ ] Change link styling to:
```
a {
    border-bottom: 1px solid transparent;
    color: #006CB8;
    text-decoration: none;
}
a:hover, a:focus, a:active {
    border-bottom: 1px solid #00548f;
    color: #00548f;
    text-decoration: none;
}
```

Before
Screen Shot 2020-07-02 at 1 28 53 PM

After
Screen Shot 2020-07-02 at 1 11 38 PM

Note that the bottom border in hover state won't appear on the titles in search results until that display is fixed. #687
Link text colour and border should also apply to `.page-link`, but not to other button-styled links.

### 2. The Bootstrap blue button does not provide enough contrast.

- [ ] Use the link text colour `color: #006CB8;` as the background, and the hover colour `color: #00548f;` for ... hover.

Before: Contrast ratio 3.97:1
Screen Shot 2020-07-02 at 1 22 26 PM

After: Contrast ratio 5.47:1
Screen Shot 2020-07-02 at 1 21 46 PM

### 3. Selected facet header does not provide enough contrast
This is the weirdest thing. `#212529` text on `#009b76` background has a contrast ratio of 4.37:1. Target is 4.5:1
Screen Shot 2020-07-02 at 2 53 59 PM

Obvious choice would be to switch the text to white
Screen Shot 2020-07-02 at 2 54 28 PM

Clearly better, right? But the contrast ratio is now 3.52:1.

I tested both versions with a vision simulator, and with every type of colour-blindness and degree of contrast loss, the white is more perceptible. Tried a real life test with a colour-blind low-vision human with same result: "The white letters leap out from the dark background."

- [ ] Change text colour to `#ffffff` on `#009b76` background, with the understanding that it will continue to be reported as an error.

jkeck:
…text-decoration.

Closes #693 

This is appears to be the last part of #693.  The link color had already been addressed.

## Before
link-styling-before

## After
link-style-after

### DIFF
--- a/app/assets/stylesheets/modules/base_styles.scss
+++ b/app/assets/stylesheets/modules/base_styles.scss
@@ -1,3 +1,15 @@
 body {
   overflow-x: hidden;
 }
+
+a {
+  border-bottom: 1px solid transparent;
+  text-decoration: none;
+
+  &:active,
+  &:focus,
+  &:hover {
+    border-bottom: 1px solid $link-hover-color;
+    text-decoration: none;
+  }
+}


app/assets/stylesheets/modules/base_styles.scss
```diff
@@ -1,3 +1,15 @@
 body {
   overflow-x: hidden;
 }
+
+a {
```
cbeer, line 5: I'll take this for a spin tomorrow; I remember trying this and ended up with borders where they didn't belong (like in the global footer, and other oddities)

jkeck: Oh, heh, good point, and it looks like we are getting some of those oddities in SearchWorks currently. 😅

jkeck:
Will let @cbeer take a look tomorrow (happy to pair up on that).  We'll need to get more clever.

PR #725, opened 2020-07-14
Style links to use a bottom border instead of text-decoration on hover.

### LINKED ISSUE #693: Contrast and identity colour changes for links, buttons, text

jkeck:
This also forces no bottom border in a few places we do not want it
(around branding images) and Bootstrap would handle for us with text-decoration.

Another try to close #693 

I did realize that the [spec](https://www.w3.org/TR/CSS21/box.html#border-color-properties) indicates that if a border is applied w/o a color specified it uses the color attribute of that element, so that is helpful. 

This does end up needing us to remove the border in a few place (which is kind of a pain). 

One place where this doesn't quite look right is in the tools menu because of the padding that Bootstrap applies to list group items.
Screen Shot 2020-07-14 at 4 00 38 PM

This looks to be all the other links in the site pretty well (but somebody should definitely take this for a spin and see if I'm missing something or should take a diff. approach).

### DIFF
--- a/app/assets/stylesheets/modules/base_styles.scss
+++ b/app/assets/stylesheets/modules/base_styles.scss
@@ -1,3 +1,15 @@
 body {
   overflow-x: hidden;
 }
+
+a {
+  border-bottom: 1px solid transparent;
+  text-decoration: none;
+
+  &:active,
+  &:focus,
+  &:hover {
+    border-bottom: solid 1px;
+    text-decoration: none !important;
+  }
+}
--- a/app/assets/stylesheets/modules/header_navbar.scss
+++ b/app/assets/stylesheets/modules/header_navbar.scss
@@ -8,6 +8,10 @@
     height: auto;
     line-height: 30px;
     width: 175px;
+
+    &:hover {
+      border-bottom: 1px solid transparent;
+    }
   }
 }
 
--- a/app/assets/stylesheets/modules/sul_footer.scss
+++ b/app/assets/stylesheets/modules/sul_footer.scss
@@ -54,6 +54,10 @@ $sul-footer-height: 80px;
   display: table-cell;
   vertical-align: middle;
   width: 162px;
+
+  a:hover {
+    border-bottom: 1px solid transparent;
+  }
 }
 
 #sul-footer-links {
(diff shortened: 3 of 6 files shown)


thatbudakguy:
Closing because I don't think we want to solve this problem this way (and modern bootstrap might solve it for us).

PR #729, opened 2020-07-29
Update to rails 6

jkeck:


config/application.rb
```diff
@@ -9,7 +9,8 @@
 module Earthworks
   class Application < Rails::Application
     # Initialize configuration defaults for originally generated Rails version.
-    config.load_defaults 5.1
+    config.load_defaults 6.0
+    config.autoloader = :classic # Our helper overrides don't work under the classic loader
```
jkeck, line 13: Was really surprised about this, but I can confirm that the zeitwerk code loader appears to break how we override helpers.  Not sure if this is specific to EarthWorks or if we'll need to keep an eye out for this in other projects.

mejackreed: huh.. thanks for pointing that out. We are probably at a good place if we need to change anything up in GeoBlacklight for a major release.

jkeck: Thanks for writing the test that failed on the upgrade!  😆

PR #733, opened 2020-08-10
Address upstream GeoBlacklight changes

jkeck:
The icon changes for the zero results page seem like they might be the most attention as they are significantly different.

## Zero Results Page (before)
zero-results-before

## Zero Results Page (after)
zero-results-after

---

## Unavailable Document (before)
unavailable-doc-before

## Unavailable Document (afteer)
unavailable-doc-after

### DIFF
--- a/app/views/catalog/_zero_results.html.erb
+++ b/app/views/catalog/_zero_results.html.erb
@@ -1,10 +1,10 @@
 <h2><%= t 'blacklight.search.zero_results.title' %></h2>
 <div id="documents" class="noresults">
-  <h3 class='modify-search'> <%= t 'blacklight.search.zero_results.modify_search' %></h3>
+  <h3 class='modify-search'><%= blacklight_icon 'earthworks-search' %> <%= t 'blacklight.search.zero_results.modify_search' %></h3>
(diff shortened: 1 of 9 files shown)


app/views/catalog/_zero_results.html.erb
```diff
@@ -1,12 +1,12 @@
 <h2><%= t 'blacklight.search.zero_results.title' %></h2>
 <div id="documents" class="noresults">
-  <h3 class='modify-search'> <%= t 'blacklight.search.zero_results.modify_search' %></h3>
+  <h3 class='modify-search'><%= blacklight_icon 'earthworks-search' %> <%= t 'blacklight.search.zero_results.modify_search' %></h3>
```
jkeck, line 3: I tried to use blacklight's search icon, but. could not get it to resize at all.  If somebody knows what I was doing wrong, happy to yank this icon for the Blacklight provided one.

jkeck:
Oh, and I guess I should have prob. read through my notifications before doing this, because I see #731 😅 

That PR doesn't address the icons on the zero results page (but maybe that's fine).

PR #768, opened 2021-02-09
Update to Rails 6.1

mejackreed:


config/application.rb
```diff
@@ -9,7 +9,7 @@
 module Earthworks
   class Application < Rails::Application
     # Initialize configuration defaults for originally generated Rails version.
-    config.load_defaults 6.0
+    config.load_defaults 6.1
     config.autoloader = :classic # Our helper overrides don't work under the new loader
```
mejackreed, line 13: Unfortunately we still can't use zeitwerk.. 🤷

mejackreed: Upstream issue filed https://github.com/geoblacklight/geoblacklight/issues/996 . Seems like we need to modify how things behave are loaded up there to make the magic happen here.

cbeer:
Should we bump some of the other dependencies to their Rails 6-generated versions?

```
# Use sqlite3 as the database for Active Record
gem 'sqlite3', '~> 1.4'
# Use Puma as the app server
gem 'puma', '~> 5.0'
# Use SCSS for stylesheets
gem 'sass-rails', '>= 6'
```

and, replacing the development block:

```

group :development do
  # Access an interactive console on exception pages or by calling 'console' anywhere in the code.
  gem 'web-console', '>= 4.1.0'
  # Display performance information such as SQL time and flame graphs for each request in your browser.
  # Can be configured to work on production as well see: https://github.com/MiniProfiler/rack-mini-profiler/blob/master/README.md
  gem 'rack-mini-profiler', '~> 2.0'
  gem 'listen', '~> 3.3'
  # Spring speeds up development by keeping your application running in the background. Read more: https://github.com/rails/spring
  gem 'spring'
end

```

🤷‍♂️ ?

mejackreed:
Great idea, I've added these.

PR #799, opened 2021-06-15
Display pre-indexed stanford rights information

cbeer:
These should appear the same way they have when rendering out of rightsMetadata, except we're dropping the license badge and linking out to the full license to bring it in line with our other apps.
 
Screen Shot 2021-06-15 at 10 26 33


app/views/catalog/_rights_metadata.html.erb
```diff
@@ -3,17 +3,28 @@
...
+  <% elsif rights_metadata.copyright %>
     <dt class='col-md-3'>Copyright</dt>
     <dd itemprop='copyrightHolder'  class='col-md-9'><%= rights_metadata.copyright %></dd>
   <% end %>
 
-  <% if rights_metadata.license %>
+  <% if document.first('stanford_license_s') %>
+    <dt class='col-md-3'>License</dt>
+    <dd itemprop='license' class='col-md-9'>
+      <%= document.first('stanford_license_s').html_safe %>
```
mjgiarlo, line 25: @cbeer can you say more about why the license is marked as safe but copyright and use statement aren't?

mjgiarlo (approved):
Have a minor question but am approving nonetheless.

app/views/catalog/_rights_metadata.html.erb
cbeer: mods_display puts some HTML in there 😄

PR #812, opened 2021-09-06
Update dependencies

sul-devops-team:


cbeer:
It's probably worth spinning the app up and poking around a little before merging this one.

ggeisler:
@mjgiarlo That increase in icon size looks fine to me (wouldn't want much bigger but they don't seem too big to me).

PR #871, opened 2022-06-07
Use redis as the cache store

thatbudakguy:
- Add hiredis gem for faster redis connections
- Use redis as the cache store in production

### DIFF
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -55,8 +55,8 @@ Rails.application.configure do
   # Prepend all log lines with the following tags.
   config.log_tags = [ :request_id ]
 
-  # Use a different cache store in production.
-  # config.cache_store = :mem_cache_store
+  # Use redis as the cache in production.
+  config.cache_store = :redis_cache_store, { url: ENV['REDIS_URL'] || Settings.REDIS_URL }
 
   # Use a real queuing backend for Active Job (and separate queues per environment).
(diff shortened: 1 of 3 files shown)


thatbudakguy:
this branch is currently deployed to staging: https://earthworks-stage.stanford.edu/

config/environments/production.rb
```diff
@@ -55,8 +55,8 @@
   # Prepend all log lines with the following tags.
   config.log_tags = [ :request_id ]
 
-  # Use a different cache store in production.
-  # config.cache_store = :mem_cache_store
+  # Use redis as the cache in production.
+  config.cache_store = :redis_cache_store, { url: ENV['REDIS_URL'] }
```
cbeer, line 59:
```suggestion
  config.cache_store = :redis_cache_store, { url: Settings.REDIS_URL }
```

Is there magic I'm missing, or should we use the railsconfig configuration from over in shared_configs?

thatbudakguy: I saw `ENV['REDIS_URL']` in the sidekiq initializers so figured that was the best thing to do, although I guess I didn't look super hard since it does both of those things:

https://github.com/sul-dlss/earthworks/blob/1a04d54c3bfec0a290deb1f4f00e2707ab5684a4/config/initializers/sidekiq.rb#L1-L7

Should we just do the same here?

cbeer: 👍

PR #876, opened 2022-06-20
Docker dev setup

thatbudakguy:
- Add docker setup for development
- Use production sidekiq setup if REDIS_URL is set
- Update developer docs in README
- Gitignore JavaScript files


corylown (approved):
Looks good to me. Tried it out and was up and running very quickly.

PR #963, opened 2023-03-29
Upgrade to geoblacklight 4.0

jcoyne:
Blocked by changing all our metadata to the new schema used by v4: https://opengeometadata.org/docs/about-ogm-aardvark


thatbudakguy:
# Todo

- [ ] Rework the featured items on the homepage using new View Components for this purpose
- [ ] Figure out why default leaflet zoom is wrong on show page?
- [ ] Figure out how/whether we expect to crosswalk custom Stanford rights fields to Aardvark
- [ ] Figure out why GeoMonitor CheckLayerJob spec is failing

app/controllers/catalog_controller.rb
```diff
@@ -280,17 +309,15 @@ class CatalogController < ApplicationController
     config.add_show_tools_partial(:sms, if: :render_sms_action?, callback: :sms_action, validator: :validate_sms_params)
 
     # Custom tools for GeoBlacklight
-    # rubocop:disable Layout/LineLength
-    config.add_show_tools_partial :web_services, if: proc { |_context, _config, options|
-      options[:document] && (Settings.WEBSERVICES_SHOWN & options[:document].references.refs.map { |r| r.type.to_s }).any?
-    }
```
dnoneill, line 286: Why is this removed?

Gemfile
```diff
@@ -83,7 +83,9 @@ gem 'newrelic_rpm'
 gem 'twitter-typeahead-rails'
 gem 'blacklight_range_limit', '~> 7.0'
 gem 'redis', '~> 5.0'
-gem 'geo_monitor', '~> 0.7', github: 'geoblacklight/geo_monitor'
+# Not compatible with GeoBlacklight 4.x
+# https://github.com/geoblacklight/geo_monitor/issues/12
+# gem 'geo_monitor', '~> 0.7', github: 'geoblacklight/geo_monitor'
```
hudajkhan, line 88: What is our plan in general for GeoMonitor once this upgrade is in place? Will we need to work out a way later to use it, or will be table this for good? Thanks

app/jobs/check_layer_job.rb
```diff
@@ -3,6 +3,7 @@ class CheckLayerJob < ApplicationJob
 
   ##
   # @param [GeoMonitor::Layer] layer
+  # TODO: migrate this to the geo_monitor gem itself?
```
hudajkhan, line 6: Is this still a to do? Thanks

Gemfile
hudajkhan: I see this down below https://github.com/geoblacklight/geo_monitor/issues/12 so I think that answers that question : )

app/controllers/catalog_controller.rb
hudajkhan: Seems like maybe the web services partial is now added by default in _show_sidebar.html.erb in GeoBlacklight so it no longer needs to be added a show tools partial separately.  (Seems that way looking at GeoBlacklight itself)

hudajkhan:
Looks good in general to me. @dnoneill and I have a few questions.  Since I've seen this branch functioning pretty well on staging already, happy to approve although I think we will need to hold off on merging because of indexing, etc.

hudajkhan (approved):

app/controllers/catalog_controller.rb
thatbudakguy: Yeah, we don't need to explicitly add this partial any longer

app/jobs/check_layer_job.rb
thatbudakguy: It is; I added it as https://github.com/geoblacklight/geo_monitor/issues/17 but left a reminder here for later. Something to do after GBLv4 compatibility.

PR #964, opened 2023-04-03
Harvesting updates

thatbudakguy:
- Update GeoCombine from GitHub
- Ensure OGM_REPOS config is passed to harvester
- Remove George Mason records from harvesting allowlist


jcoyne (approved):
@thatbudakguy approved. But, in the future, I think it's easier to review changes if you make only a single change per PR (and multiple PRs) rather than lump them together as "Some updates".

PR #967, opened 2023-04-06
Update GeoCombine to v0.8.0

thatbudakguy:
so we can stop pinning it to github branch. once updated upstream in geoblacklight, we no longer need to specify it here


thatbudakguy:
superseded by #969

PR #970, opened 2023-04-07
Pin faraday >= 2

thatbudakguy:
This removes the downstream dependency on patron, which needs
cURL headers in order to build correctly on Ubuntu machines.

The missing headers cause issues on both GitHub CI runners (https://github.com/actions/runner-images/issues/37)
and our automated deploys
(https://sul-ci-prod.stanford.edu/job/SUL-DLSS/job/earthworks/job/master/56/display/redirect).

### DIFF
--- a/.github/workflows/ruby.yml
+++ b/.github/workflows/ruby.yml
@@ -14,10 +14,6 @@ jobs:
         ruby: ['3.1', '3.2']
     steps:
     - uses: actions/checkout@v3
-    # required to avoid https://github.com/actions/runner-images/issues/37
-    # because faraday depends on patron, which requires curl headers to build
-    - name: Install cURL Headers
-      run: sudo apt-get update && sudo apt-get install libcurl4-openssl-dev
     - name: Set up Ruby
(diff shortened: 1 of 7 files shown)


thatbudakguy:
blocked on #969 which includes faraday 2 compatibility

cbeer:
I think code like:
https://github.com/sul-dlss/earthworks/blob/629a5dd2c336dece626baa751f829489a06034f8/scripts/geoserver/update_caching.rb#L18

isn't 2.x compatible, is it (also used in other `scripts` files)?

thatbudakguy:
Hmm, I didn't know about any of those files in `scripts`. From a cursory glance, it seems like they are similar to tasks that @kimdurante does manually right now (like updating GeoWebCache for a layer). The OGP related stuff we no longer need, since we're not committing to maintain that old data anymore (per discussion with GIS team).

I could see about making them faraday 2.x compatible if it's just a matter of changing those `basic_auth` calls. Do you know anything else about these scripts?

PR #973, opened 2023-04-10
Prevent multiple worlds being shown on the leaflet map

### LINKED ISSUE #911: Set zoom level on landing page maps so that only one world map is shown
`As someone who cares about cartographic presentation, I would like the default map to display a single instance of the world, by default.` 

This drives me crazy:

Image

in Leaflet JS, this would be fixed with:
`setMaxBounds`
and for animation:
`maxBoundsViscosity`

thatbudakguy:
Fixes #911

# before

Screenshot 2023-04-10 at 13 32 26

# after

Screenshot 2023-04-10 at 13 30 11


hudajkhan:
Additionally, this fix also addresses the map wrapping on the search results page when no bounding box has been selected and the search results on the page cover the world map. 

After this fix, the search results view in the situation where the results on the page cover the world will look like
Screenshot 2023-04-10 at 6 27 24 PM

PR #1000, opened 2023-10-05
Move search icon to a component

jcoyne:
This removes the dependency on the Redis cache, which seems to fail regularly with EOF


cbeer:
Is this just magically used, or do we need to do any sort of configuration?

jcoyne:
@cbeer it's magically used by default. If this isn't found it falls back to the old way.

PR #1041, opened 2024-06-05
Decouple CD for stage/UAT and connect to Geoblacklight v4 branch

thatbudakguy:
This will allow iterative refinement of the Geoblacklight-4
compatible version of earthworks without needing to pause
dependency updates in production.

It's expected that this work will be finished and reverted by
the start of the Earthworks workcycle on 6/24.

If this is merged, https://github.com/sul-dlss/shared_configs/pull/2194
and https://github.com/sul-dlss/shared_configs/pull/2193 are
also needed to point the app at new compatible solr cores using
the Aardvark schema.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -12,7 +12,7 @@ pipeline {
       }
 
       when {
-        branch 'main'
+        branch 'geobl4'
       }
 
       steps {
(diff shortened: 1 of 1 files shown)


cbeer:
Do you also need to tell cap which branch to deploy, or is it getting that magically?

thatbudakguy:
Good catch, thanks! I think this is the right way?

PR #1077, opened 2024-07-10
Update name of solr-core in docker

peetucket:
Solr core names differ from solr-wrapper to Docker.  This may reduce some confusion.

This makes them the same so the README instructions are correct.

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,11 +38,8 @@ Most other commands can be run by prepending `docker compose exec app` to the co
 ### Adding data
 To add a small amount of test records to the Solr index, you can use the `seed` task:
 ```sh
-SOLR_URL=http://127.0.0.1:8983/solr/earthworks bin/rake geoblacklight:solr:seed
+bin/rake geoblacklight:solr:seed
 ```
-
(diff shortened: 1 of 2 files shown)


thatbudakguy:
can we just change https://github.com/sul-dlss/earthworks/blob/264feeb834b84de5226b0d2f80c2a0f3d7f6690b/docker-compose.yml#L45

to `blacklight-core` ? I think that's pretty standard among blacklight apps and then there wouldn't be any confusion

peetucket:
> can we just change
> 
> https://github.com/sul-dlss/earthworks/blob/264feeb834b84de5226b0d2f80c2a0f3d7f6690b/docker-compose.yml#L45
> 
> to `blacklight-core` ? I think that's pretty standard among blacklight apps and then there wouldn't be any confusion

maybe, happy to experiment with this (not really clear what `entrypoint` config means and if there are other bits that would need to be updated

hudajkhan:
There is also this line here: https://github.com/sul-dlss/earthworks/blob/main/docker-compose.yml#L14 . I think it is fine for this to go in either direction (i.e. changing the docker solr or updating the readme with the instructions in the PR), as long as the instructions are clear.

peetucket:
OK, seemed to work to just change the core name.  Brought it down, brought it backup, tried the fixture create and it seemed to work.  Amending this PR to do this and remove the change to the readme.

hudajkhan (approved):

PR #1082, opened 2024-07-11
Add searchworks url to toolbar

### LINKED ISSUE #310: Add link to view in SearchWorks
Potentially could be troublesome as records can be unreleased

peetucket:
Fixes #310 and follow on from https://github.com/sul-dlss/searchworks_traject_indexer/pull/1490 

If searchworks url is present in solr document, render on the page.

Todo:
- [ ] styling
- [ ] move location on the page?
- [x] use latest geoblacklight (main)?

With link:

http://localhost:3000/catalog/stanford-dt131hw5005

Screenshot 2024-07-11 at 1 57 28 PM

### DIFF
--- a/app/controllers/catalog_controller.rb
+++ b/app/controllers/catalog_controller.rb
@@ -309,15 +309,12 @@ class CatalogController < ApplicationController
     config.add_show_tools_partial(:sms, if: :render_sms_action?, callback: :sms_action, validator: :validate_sms_params)
 
     # Custom tools for GeoBlacklight
-    config.add_show_tools_partial :metadata, if: proc { |_context, _config, options|
-      options[:document] && (Settings.METADATA_SHOWN & options[:document].references.refs.map { |r| r.type.to_s }).any?
-    }
-    config.add_show_tools_partial :arcgis, partial: 'arcgis', if: proc { |_context, _config, options|
-      options[:document] && options[:document].arcgis_urls.present?
-    }
-    config.add_show_tools_partial :data_dictionary, partial: 'data_dictionary', if: proc { |_context, _config, options|
-      options[:document] && options[:document].data_dictionary_download.present?
-    }
+    config.add_show_tools_partial :searchworks_url, component: Earthworks::SearchworksUrl,
+                                                    if: proc { |_context, _config, options|
+                                                          options[:document] &&
+                                                            options[:document].searchworks_url.present?
+                                                        }
+
     config.show.document_actions.delete(:sms)
 
     # Configure basemap provider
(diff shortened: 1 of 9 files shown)


app/controllers/catalog_controller.rb
```diff
@@ -309,15 +309,12 @@ class CatalogController < ApplicationController
...
     # Custom tools for GeoBlacklight
-    config.add_show_tools_partial :metadata, if: proc { |_context, _config, options|
-      options[:document] && (Settings.METADATA_SHOWN & options[:document].references.refs.map { |r| r.type.to_s }).any?
-    }
-    config.add_show_tools_partial :arcgis, partial: 'arcgis', if: proc { |_context, _config, options|
-      options[:document] && options[:document].arcgis_urls.present?
-    }
-    config.add_show_tools_partial :data_dictionary, partial: 'data_dictionary', if: proc { |_context, _config, options|
-      options[:document] && options[:document].data_dictionary_download.present?
-    }
```
peetucket, line 320: these can now all come from geoblacklight components

app/helpers/earthworks_geoblacklight_helper.rb
```diff
@@ -1,9 +1,8 @@
 module EarthworksGeoblacklightHelper
   include GeoblacklightHelper
 
-  def document_available?
-    (@document.public? && @document.available?) ||
-      (@document.same_institution? && user_signed_in? && @document.available?)
+  def document_available?(document = @document)
+    super && document.available?
   end
```
peetucket, line 6: because of other changes made in `item_map_viewer_component`...see https://stanfordlib.slack.com/archives/C0AAK1HNY/p1720731983010169

app/controllers/catalog_controller.rb
dnoneill: don't them need to be updated, not removed?

peetucket: to me the conditional logic for when to show the tools to me looks identical in geoblacklight, and i couldn't even find the partials that are referenced here in earthworks ... so I was presuming that since earthworks is derived from geoblacklight, they don't need to be re-defined here 

but i'm not 100% sure, would certainly be nice to validate with enough real data to be confident

hudajkhan: Checking core GeoBlacklight code, the ArcGIS and DataDictionary components are now included.  I also see the metadata partial being added similar to the above, so including the latest GeoBlacklight code on the main branch means we no longer require the additions above. For reference: https://github.com/geoblacklight/geoblacklight/blob/main/lib/generators/geoblacklight/templates/catalog_controller.rb#L309

hudajkhan (approved):

Issue #41, opened 2014-11-13
add synonyms to Solr index for better searching results


drh-stanford:
OGP has synonym files that handled LCSH and more.

https://github.com/OpenGeoportal/ogpSolrConfig/tree/master/ogpSolrConfig/SolrConfig

thatbudakguy:
work to be done here:

- copy the synonym .txt files from the link at top into the `config/solr_config` directory of this repo
- update the `schema.xml` and/or `solrconfig.xml` in this repo to make the analyzers take account of the synonym files
- test it out locally to see if synonym-based searching works (e.g. "CO" or "colo" should give results for "Colorado")
- if it works, make the same changes to the solr config in `sul-solr-configs`
- do a reindex, if needed? (might not be needed if we only apply synonyms at query time)
- consider making the same changes in geoblacklight

hudajkhan:
The solr configuration and schema files in the repository are already out of sync with our latest changes to Aardvark.  As part of this PR, I will also update the files to reflect what we have in sul-solr-configs before adding the changes.

PR #1083, opened 2024-07-11
Adding synonym values based on OGP examples

hudajkhan:
Addresses #41 .
**What this pull request does:**
Copies the additional synonym lines from the following files under the originally referenced OpenGeoPortal configuration: 
https://github.com/OpenGeoportal/ogpSolrConfig/tree/master/ogpSolrConfig/SolrConfig

These include synonyms for LCSH terms (e.g. mapping "mountains" and "peaks" to each other so that any query for mountains would include both mountains and peaks in the query after synonyms have been applied), synonyms for ISO  (e.g. mapping "History" and "ancestry" to each other), and State names (e.g. mapping Wyoming and "WY " - with a trailing space - to each other). 

An additional pull request will apply these changes to sul-solr-configs starting with stage. 

**Using synonyms in Solr**
One of the LCSH mappings is between mountains and peaks (and several other words). 

There is a Solr record "HOV Lanes, San Francisco Bay Area, 2002" which contains the word "peak" but not "mountains" in its fields. 

On my local machine, I have an example of an EarthWorks solr index with this document which uses the current schema/solrconfig and another example which uses a version with the updated synonyms file. 

Using the non-updated EarthWorks solr, when querying for the word "mountains", we do not see this document in the result.  The analysis page also shows no synonyms are in play. 

Using the updated EarthWorks solr, when querying for the word "mountains", we do see this document in the results.  The analysis page also shows the synonyms for "mountains" is in play. 

Screenshot 2024-07-11 at 3 17 13 PM

Part of the analysis page is below (which shows the other synonyms in the list for "mountains"):
Screenshot 2024-07-11 at 3 18 36 PM

**Additional context from the OGP side:**

OpenGeoPortal had separate fields for themes, place names, and layer display names (in addition to others).  They also had versions of these fields which also used field types that relied on the appropriate synonym file e.g. a theme field that uses the LCSH synonyms.  They could then utilize different boosting for these fields: e.g. https://github.com/OpenGeoportal/OGP2/blob/master/geoportal_1/src/main/webapp/resources/javascript/lib/solr.js#L318 .  

In our case, we are adding all the synonym files directly for the "text_en" field type i.e. the "text" field that is used as part of regular queries (for reference, see the /select handler and qf and pf which both employ the text field).

### DIFF
--- a/config/solr_configs/synonyms.txt
+++ b/config/solr_configs/synonyms.txt
@@ -33,117 +33,119 @@ Television, Televisions, TV, TVs
 # Synonym mappings can be used for spelling correction too
 pixima => pixma
 
-# Copied from OGP LCSH synonyms
+# Copied from OGP LCSH synonyms, with duplicates removed from multi-way expansion synonyms 
 Harbors,Pier,Quay,Warf
 
 Hydrocarbons,CNG,Coal,ethanol,gas,gasoline,kerosene,LNG,oil,OPEC,petrol,Petroleum,propane
 
-Walls,Fences,Fortification,fortresses,Gates,Hedges,Seawall,Wall,Walls
+Walls,Fences,Fortification,fortresses,Gates,Hedges,Seawall,Wall
 
-Roads,arteries,automobiles,Avenue,Avenues,beltways,Boulevards,driveways,expressway,expressways,Freeways,highway,highways,intersections,interstate,interstates,lanes,motorways,road,Roadblocks,Roads,Roadways,route,Routes,Speedlimits,Street,Streets,Thoroughfares,Tollbooths,way
+Roads,arteries,automobiles,Avenue,Avenues,beltways,Boulevards,driveways,expressway,expressways,Freeways,highway,highways,intersections,interstate,interstates,lanes,motorways,road,Roadblocks,Roadways,route,Routes,Speedlimits,Street,Streets,Thoroughfares,Tollbooths,way
 
-Schools,academies,College,colleges,colleges/universities,Education,preschool,preschools,school,Schools,universities,university,University/Colleges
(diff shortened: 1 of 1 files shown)


hudajkhan:
I have made some changes. See #1091 .

Specifically - a lot of the multi-expansion lines (i.e. any word on a line - if it's in a query - will have all the other words on the line also added to the query), had lots of duplicates. E.g. Elevation, altitudes, elevation.  I removed the duplicate words from many of these lines.  Since our configuration ignores case, I also removed duplicates that 

I also saw "buildings" in the multi-word expansion in a line about religion. That didn't make much sense so I removed the word "buildings". 

And yes, for the lobsters line, I replaced "Invertebrates, lobsters,mosquitoes,shellfish" with "Invertebrates => Invertebrates,lobsters,mosquitoes,shellfish".  This expands a query for invertebrates to also include lobsters, mosquitoes, and shellfish, but not the other way around.

config/solr_configs/synonyms.txt
```diff
@@ -18,6 +18,10 @@ ccc => cccc1,cccc2
 a\=>a => b\=>b
 a\,a => b\,b
 fooaaa,baraaa,bazaaa
+#The OGP file also includes these mappings
+aaafoo => aaabar
+bbbfoo => bbbfoo bbbbar
+cccfoo => cccbar cccbaz
```
thatbudakguy, line 24: pretty sure these are fake/for testing?

hudajkhan: Yes, and they seem to extend the mappings for testing that are at the very beginning of our own file.

thatbudakguy: I guess no harm keeping them in for testing 🤷🏻

thatbudakguy (approved):

PR #1084, opened 2024-07-12
move arcgis link to stanford version

### LINKED ISSUE #912: Add "Open in ArcGIS.com" Button
`As a researcher using spatial data, I want to open the dataset I have found in Earthworks, directly in stanford.maps.arcgis.com, so that I can begin using it. `

Or, perhaps, "Send to ArcGIS.com"

Similar in spirit to Open in CARTO, but our relationship with Esri is more stable. Our Esri site license is considered infrastructure, so will never go away. This would create the ability to add WFS/WMS sources, directly to ArcGIS Online, where we currently have 2000+ users. 

Possibly can be done through the APIs? 

https://developers.arcgis.com/documentation/mapping-apis-and-services/data-hosting/data-publishing/#fn:support-table-1:2

We can call on Esri for collaboration with their developers on this.

dnoneill:
closes #912

### DIFF
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -1,5 +1,5 @@
 # ArcGIS Online Base URL
-ARCGIS_BASE_URL: "https://www.arcgis.com/home/webmap/viewer.html"
+ARCGIS_BASE_URL: "https://stanford.maps.arcgis.com/home/webmap/viewer.html"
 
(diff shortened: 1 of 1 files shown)


dnoneill:
New designs remove arcgis link

PR #1086, opened 2024-07-12
rake task and cron job for clearing stale downloads

### LINKED ISSUE #205: add service for cleaning cache periodically
through some "smart" means -- LRU by downloads...

peetucket:
Fixes #205 -- add a scheduled rake task for clearing stale download files

Configured to run daily and clear any files not accessed in the last 2 weeks (time period configurable in settings).

Tested find command manually on server (but nothing deleted).

~~See also https://github.com/sul-dlss/shared_configs/pull/2220 which is required first.~~

~~Note, we may want to set the download cache directory in stage shared configs here: https://github.com/sul-dlss/shared_configs/blob/earthworks-stage/config/settings/production.yml#L1  to match what is in in prod: https://github.com/sul-dlss/shared_configs/blob/earthworks-prod/config/settings/production.yml#L1~~

~~it is currently set to `nil` in stage, which will cause problems with this cron job~~

### DIFF
--- a/config/initializers/okcomputer.rb
+++ b/config/initializers/okcomputer.rb
@@ -6,9 +6,7 @@ OkComputer::Registry.register 'solr',
   OkComputer::HttpCheck.new(Blacklight.default_index.connection.uri.to_s.sub(%r{/$}, '') + '/admin/ping')
 
 OkComputer::Registry.register 'downloads-cache',
-  OkComputer::DirectoryCheck.new(
-    Settings.DOWNLOAD_PATH || (Rails.root + 'tmp/cache/downloads').to_s, true
-  )
+  OkComputer::DirectoryCheck.new(Settings.DOWNLOAD_PATH, true)
 
 OkComputer::Registry.register 'redis',
   OkComputer::RedisCheck.new(url: ENV.fetch('REDIS_URL') { Settings.REDIS_URL })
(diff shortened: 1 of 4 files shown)


config/settings.yml
```diff
@@ -2,7 +2,8 @@
 ARCGIS_BASE_URL: "https://www.arcgis.com/home/webmap/viewer.html"
 
 # DOWNLOAD_PATH is set in production but defaults to this tmp dir when not set
-# DOWNLOAD_PATH: './tmp/cache/downloads'
+DOWNLOAD_PATH: './tmp/cache/downloads'
+download_cache_expiry_time_days: 14 # the number of days since a download was last accessed before it is removed from the cache
```
peetucket, line 7: new setting for expiry time and explicitly make it clear what the download path is in development (it is overriden by shared configs for all deployed environments)

config/initializers/okcomputer.rb
```diff
@@ -6,9 +6,7 @@
   OkComputer::HttpCheck.new(Blacklight.default_index.connection.uri.to_s.sub(%r{/$}, '') + '/admin/ping')
 
 OkComputer::Registry.register 'downloads-cache',
-  OkComputer::DirectoryCheck.new(
-    Settings.DOWNLOAD_PATH || (Rails.root + 'tmp/cache/downloads').to_s, true
-  )
+  OkComputer::DirectoryCheck.new(Settings.DOWNLOAD_PATH, true)
```
peetucket, line 10: the `DOWNLOAD_PATH` should always exist now -- either the default in `settings.yml` below or the actual value from shared_configs

thatbudakguy (approved):

PR #1097, opened 2024-07-24
Drop app container from Docker setup

### LINKED ISSUE #1037: Consider changing docker-compose to only run Solr and Postgres, not the Rails app
See for example our setup in Stanford-Arclight: https://github.com/sul-dlss/stanford-arclight/blob/fa8810e684150f82e676fd3279c1ab159900b6cc/docker-compose.yml -- this was a very handy way to work, it was basically a swap out for `solr_wrapper` with Docker.

In this approach, Rails app still runs directly on the local machine, so you do not need to prefix commands with `docker compose exec app`. You are also able to run tests successfully. Since we aren't deploying this app via Docker, removing it from the docker setup is a more straightforward dev setup in my opinion. 

Writing up this ticket after a discussion with Nick and some minor work in #876

jmartin-sul:
* drop `app` container from `docker-compose.yml`
* drop `Dockerfile` and `docker-entrypoint.sh` files entirely since those were only there to support the `app` container
* add the dotenv gem as a dependency, and add [`.env` and `.env.test` files](https://github.com/bkeepers/dotenv/tree/b86d9e4c23a9f1bc83f8853f6a1b244449c65cd0?tab=readme-ov-file#customizing-rails) that can be uncommented to leverage the Docker-ized stack
  * As far as I could tell from the README and the `database.yml`, there wasn't already a documented approach for pointing Rails server running directly on the laptop to the containerized PG instance or to Redis (makes sense since the app container that's dropped by this PR had those connections).  So the new dependency seemed worth it for not having to prefix multiple commands with a `DATABASE_URL` or other env vars -- happy to drop the second commit or otherwise revisit if there's a better approach I didn't think of.
  * If the second commit should stick around, I'm happy to squash this before merge.
* updated the README accordingly

To test, I ran through the updated instructions using both the default and Docker approaches.  Confirmed after dropping the DB and uncommenting the `.env` contents that the SQLite files didn't reappear, the app ran, the tests ran, and the ActiveRecord connection was to the Postgres container.

closes #1037


.env
```diff
@@ -0,0 +1,5 @@
+# DATABASE_URL='postgresql://earthworks:earthworks@localhost/earthworks?pool=5'
+
+# REDIS_URL: redis://localhost:6379
+# REDIS_HOST: cache
+# REDIS_PORT: 6379
```
jmartin-sul, line 5: still defaults to SQLite and no Redis

README.md
jmartin-sul, line 1: I reorganized this file a bit, and added a bit of whitespace -- all changes that felt like they made for a better step by step flow, but of course happy to revisit if others find this more confusing.

thatbudakguy (approved):
this looks great!

PR #1100, opened 2024-07-24
Code snippety

### LINKED ISSUE #1049: Add autogenerated code samples for using layers
You can view this prototype by visiting the [interactive prototype](https://www.figma.com/proto/nkgg905s5FyXoyPZjVMco0/EarthWorks?page-id=0%3A1&node-id=89-2697&viewport=-3973%2C-836%2C0.58&t=5oPw70wWwebL9uPW-1&scaling=min-zoom&starting-point-node-id=9%3A148&show-proto-sidebar=1), visiting a layer detail page, and clicking "code samples" in the tools panel.

This deserves a little bit of thought with regard to maintainability: do we just have the code samples as string templates that get the layer ID, etc. injected into them? Do we do any work to validate that the code still runs/is up to date? Can they be tested? etc.

If directly writing out the code isn't the best way, perhaps step-by-step instructions or links would still be helpful. @mapninja is a good person to ask about how this will be used in practice (e.g. by students).

hudajkhan:
Closes #1049 .

What this pull request does:

Adds the ability to see code samples
- Adds code snippet link to show tools. (Note: the UI designs show this in a different place, so we will need to move the link to the right place once we start working on UI changes)
- Reads code snippets from static files. (Not all the files are complete, and some are placeholders).  
- Uses the following logic to determine which file needs to be displayed:  if resource type is "Raster data", show raster code samples.  If resource type is "Vector data","Polygon data", "Point data", "Line data", or "Index Maps", show vector samples.
- Show tabs for three languages: Python, R, Leaflet. 
- I've added some specific tokens for wxs identifier, geoserver wms, and geoserver wfs within some code samples, to allow these values to be replaced. This will allow the static template to say <<WXS_ID>>, for example, while the code sample page will be able to get that field from the Solr document and insert the value into the page. 
- Clicking the "Copy" button will show "Copied" with a check mark to the left.  Clicking on any of the tab links, including the current one, will change the button text back to "Copy" and hide the check mark.

### DIFF
--- a/app/assets/javascripts/modules/copy_clipboard.js
+++ b/app/assets/javascripts/modules/copy_clipboard.js
@@ -1,0 +1,71 @@
+/*
+ * This code is used by the copy to clipboard function for the code snippet modal
+*/
+const setupCopyToClipboardListener = function () {
+  const modalElement = document.getElementById("blacklight-modal");
+  
+  // For a click event on the modal itself, capture that and then check if it is the clipboard copy button
+  modalElement.addEventListener("click", function(e){
+    const button = e.target.closest(".btn-clipboard");
+    if (button) {
+      // Copy the text content from the selected pane
+      copyActivePaneContent();
+      // Change the text of the click button and add a checked icon
+      displayCopyClick();
+    }
+    const tabPane = e.target.closest(".code-snippet-tab");
+    if(tabPane) {
+      resetCopyClick();
+    }
+  })
+}
+
+// Copy active code panel's code, based on which programming language the user has selected
+const resetCopyClick = function() {
+  const copyButtonText = document.querySelector("span#copyClipText");
+  const copyCheck = document.querySelector("span#copyCheck");
+  if(! copyButtonText) {
+    console.log("Error finding copy button");
+    return;
+  }
+
+  // Check if the check mark is visible, toggle on that basis
+  if(! copyCheck.classList.contains("d-none")) {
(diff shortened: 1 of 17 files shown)


app/assets/javascripts/modules/copy_clipboard.js
```diff
@@ -0,0 +1,71 @@
+/*
+ * This code is used by the copy to clipboard function for the code snippet modal
+*/
+const setupCopyToClipboardListener = function () {
```
marlo-longley, line 4: This PR in GeoBlacklight adds a clipboard stimulus controller: https://github.com/geoblacklight/geoblacklight/pull/1441/files

It may be helpful to use it in the future.

hudajkhan: Thanks @marlo-longley ! I can look at changing this PR to the latest GBL EarthWorks branch and integrating that, if that makes more sense.

hudajkhan:
I've added @mapninja's latest examples. I believe this covers all the examples for Vector/Raster.

hudajkhan:
I am going to go ahead and close this and create a new PR against BL8 instead.

hudajkhan:
Re-opening for review, but may create a new PR based on the new BL8 branch

hudajkhan:
New PR at https://github.com/sul-dlss/earthworks/pull/1125

Issue #1053, opened 2024-06-11
Update the header


thatbudakguy:
See [the interactive prototype](https://www.figma.com/proto/nkgg905s5FyXoyPZjVMco0/EarthWorks?page-id=0%3A1&node-id=190-4386&viewport=-3973%2C-836%2C0.58&t=g6nuKFbuALQqwBG7-1&scaling=min-zoom&starting-point-node-id=190%3A4386&show-proto-sidebar=1).

This work was done recently in stanford-arclight, e.g. https://github.com/sul-dlss/stanford-arclight/blob/main/app/views/shared/_stanford_stripe.erb

thatbudakguy:
See also the reference implementation (if it's ready?) in https://github.com/sul-dlss/component-library

jcoyne:
Possibly depends on https://github.com/sul-dlss/earthworks/issues/1063

dbranchini:
Note the mobile menu in Arclight (which splits the header to display the menu) is not the final or what DLUX decided on. It should display the navigation below the EarthWorks title as seen here - 

https://www.figma.com/proto/nkgg905s5FyXoyPZjVMco0/EarthWorks?page-id=0%3A1&node-id=335-15305&viewport=-7294%2C1192%2C0.56&t=kYki5n8Z1jZKPmeu-1&scaling=min-zoom&content-scaling=fixed&starting-point-node-id=335%3A15305

PR #1104, opened 2024-07-30
update header to stanford header

dnoneill:
Screenshot 2024-08-01 at 11 56 22 AM
Screenshot 2024-08-01 at 11 51 17 AM
Screenshot 2024-08-01 at 11 51 07 AM

closes [#1053](https://github.com/sul-dlss/earthworks/issues/1053)


app/assets/stylesheets/header.scss
```diff
@@ -0,0 +1,78 @@
+.su-brand-bar__logo {
```
jcoyne, line 1: Are you able to follow the procedure laid out here https://github.com/sul-dlss/component-library?tab=readme-ov-file#releasing ?  This is what we are doing in argo and purl.  If we have a common way of doing this, then it will be easier to manage in the long run.

jcoyne (changes requested):

app/components/blacklight/top_navbar_component.html.erb
```diff
@@ -0,0 +1,192 @@
+<nav id="skip-link" role="navigation" aria-label="Skip links">
```
jcoyne, line 1: In BL, the skip-link is on the base template. https://github.com/projectblacklight/blacklight/blob/abf6f24bb8e53354dc91db8520f029bee68c4a4c/app/views/layouts/blacklight/base.html.erb#L31-L37. Can we continue to follow that paradigm here?  That makes it easier for people who navigate between the two code bases.

jcoyne (changes requested):

app/components/blacklight/top_navbar_component.html.erb
dnoneill: Done. Thanks.

app/assets/stylesheets/header.scss
dnoneill: We don't have a base.html.erb template. I am guessing we will need one eventually so I can add it in but right now I am not sure how to do this without creating the file?

app/assets/stylesheets/geoblacklight.scss
```diff
@@ -1,3 +1,7 @@
 @import 'bootstrap/scss/bootstrap';
 @import 'blacklight-frontend/app/assets/stylesheets/blacklight/blacklight';
-@import '@geoblacklight/frontend/app/assets/stylesheets/geoblacklight/geoblacklight';
+@import '@geoblacklight/frontend/dist/stylesheets/geoblacklight/geoblacklight';
```
dnoneill, line 3: This might need to change but with v3 it is what works

app/views/layouts/blacklight/base.html.erb
```diff
@@ -0,0 +1,55 @@
...
+      <%= javascript_include_tag "application", "data-turbo-track": "reload", defer: true %>
+    <% else %>
+      <%= javascript_include_tag "application", "data-turbo-track": "reload", defer: true %>
+      <%= javascript_include_tag "blacklight/blacklight", type: 'module' %>
+      <script type="module">
+        import githubAutoCompleteElement from 'https://cdn.skypack.dev/@github/auto-complete-element';
+      </script>
+    <% end %>
+
+    <link rel="stylesheet" href="/sul-components/v2024-08-01/styles/sul.css" />
```
dnoneill, line 27: This is just the blacklight base with this link added.

app/views/layouts/blacklight/base.html.erb
```diff
@@ -0,0 +1,55 @@
...
+  </head>
+  <body class="<%= render_body_class %>">
+    <nav id="skip-link" role="navigation" class="visually-hidden-focusable sr-only sr-only-focusable" aria-label="<%= t('blacklight.skip_links.label') %>">
+      <div class="container-xl">
+        <%= link_to t('blacklight.skip_links.search_field'), '#search_field', class: 'd-inline-flex p-2 m-1', data: { turbolinks: 'false' } %>
+        <%= link_to t('blacklight.skip_links.main_content'), '#main-container', class: 'd-inline-flex p-2 m-1', data: { turbolinks: 'false' } %>
+        <%= content_for(:skip_links) %>
+      </div>
+    </nav>
+    <%= render blacklight_config.header_component.new(blacklight_config: blacklight_config) %>
```
dnoneill, line 40: Only other change is using the component instead of the partial, figured it was clearer.

app/assets/stylesheets/geoblacklight.scss
```diff
@@ -1,3 +1,7 @@
...
+
+.palo-alto-base {
+  background-color: #175E54!important;
```
jcoyne, line 6: Why do we need `!important`?

jcoyne (changes requested):

app/assets/stylesheets/geoblacklight.scss
dnoneill: Yup. Bootstrap overrides the background color on the header.

jcoyne: In argo we are doing this instead:
```
.stone-dark {
  --bs-dark-rgb: var(--stanford-stone-dark-rgb);
}
```

I think you could do the same here (but for palo-alto)

dnoneill: That makes sense. Thanks. It worked.

jcoyne (approved):

app/assets/stylesheets/geoblacklight.scss
```diff
@@ -1,3 +1,7 @@
 @import 'bootstrap/scss/bootstrap';
 @import 'blacklight-frontend/app/assets/stylesheets/blacklight/blacklight';
-@import '@geoblacklight/frontend/app/assets/stylesheets/geoblacklight/geoblacklight';
+@import '@geoblacklight/frontend/dist/stylesheets/geoblacklight/geoblacklight';
+
+.palo-alto-base {
+  --bs-dark-rgb: 23, 94, 84;
+}
```
thatbudakguy, line 7: can we put this in our own `style.css`? I think this stylesheet is just for GBL stuff

dnoneill: @thatbudakguy I moved this to the application.css file in the PR for the footer.

Issue #1093, opened 2024-07-18
Explore how to incorporate authorization into code samples for restricted datasets


hudajkhan:
#1049 is about building in a code sample view.  For restricted datasets, like https://earthworks.stanford.edu/catalog/stanford-cg357zz032, the user must be logged in as a Stanford user in order to view/download the data.  

Questions to answer:

- How does EarthWorks handle authorization for the restricted GeoServer?
- Is it possible to add this authorization to the code sample?

hudajkhan:
Related references:
Proxy controller setup specifically for the restricted GeoServer URL. https://github.com/sul-dlss/earthworks/blob/main/app/controllers/restricted_proxy_controller.rb
EarthWorks shared configs has the specific restricted GeoServer URL to be used. https://github.com/sul-dlss/shared_configs/blob/earthworks-prod/config/settings/production.yml#L9
Looks like EarthWorks puppet also has related configuration. https://github.com/sul-dlss/puppet/blob/production/hieradata/node/earthworks-prod-a.stanford.edu.eyaml#L155


Related issues:
https://github.com/sul-dlss/earthworks/issues/3

marlo-longley:
Find out if this is even possible -- if so, close this exploration ticket, and create an implementation ticket that will depend on the merging of #1100

dbranchini:
Public dataset - https://www.figma.com/design/nkgg905s5FyXoyPZjVMco0/EarthWorks?node-id=911-24473&t=KfDIeJaBz5XarlOV-4

Restricted dataset - NOT LOGGED IN - https://www.figma.com/design/nkgg905s5FyXoyPZjVMco0/EarthWorks?node-id=911-24351&t=KfDIeJaBz5XarlOV-4

Restricted dataset - LOGGED IN - https://www.figma.com/design/nkgg905s5FyXoyPZjVMco0/EarthWorks?node-id=517-18605&t=KfDIeJaBz5XarlOV-4

hudajkhan:
I am going to work on adding this in to the code snippet approach in #1100 as well. I will close this issue and add these examples to #1100

marlo-longley:
See #1125

PR #1125, opened 2024-08-07
Code snippet functionality for EarthWorks

hudajkhan:
Closes https://github.com/sul-dlss/earthworks/issues/1049 .

What this pull request does:

- Add an "other options" component to the sidebar which also includes a link to open the code snippet manual
- Adds the ability to see code samples
- Adds code snippet link to show tools. (Note: the UI designs show this in a different place, so we will need to move the link to the right place once we start working on UI changes).
- Reads code snippets from static files. (Not all the files are complete, and some are placeholders).
- Uses the following logic to determine which file needs to be displayed: if resource type is "Raster data", show raster code samples. If resource type is "Vector data","Polygon data", "Point data", "Line data", or "Index Maps", show vector samples.
- Show tabs for three languages: Python, R, Leaflet.
- I've added some specific tokens for wxs identifier, geoserver wms, and geoserver wfs within some code samples, to allow these values to be replaced. This will allow the static template to say <<WXS_ID>>, for example, while the code sample page will be able to get that field from the Solr document and insert the value into the page.
- Clicking the "Copy" button will show "Copied" with a check mark to the left. Clicking on any of the tab links, including the current one, will change the button text back to "Copy" and hide the check mark.
- If the dataset is restricted, the code snippet link is not displayed at all.
- There are tests of the components, as well as one for the other options section on the page, which (currently) focuses on code snippet links displaying correctly but can also include data dictionary/metadata link checks later.


dnoneill:
@hudajkhan You have an other options component. That is what the app/components/earthworks/document/sidebar_component.rb is. It is everything in the other options. I would recommend getting rid of your other_options component and renaming the sidebar_component.

app/components/earthworks/code_snippet_modal_component.html.erb
```diff
@@ -0,0 +1,48 @@
+<%= render Blacklight::System::ModalComponent.new do |component| %>
+  <% component.with_title { t('geoblacklight.code_snippet.title') } %>
```
marlo-longley, line 2: `geoblacklight.code_snippet.title` is not coming through. Additionally, I would think this should go in the earthworks section, not geoblacklight. If you add it to `en.yml` under `earthworks`, and substitute the string location, it works for me locally:

`  <% component.with_title { t('earthworks.code_snippet.title') } %>`

app/components/earthworks/code_snippet_modal_component.rb
```diff
@@ -0,0 +1,115 @@
...
+    def render_leaflet
+      substitute_values(retrieve_text('Leaflet'))
+    end
+
+    # Incorporate
+    def substitute_values(content)
+      sub_content = content
+
+      # replace DRUID_ID
+      sub_content = sub_content.gsub('<<WXS_ID>>', @document.wxs_identifier) unless @document.wxs_identifier.empty?
```
marlo-longley, line 33: For all of these substitutions, consider handling the case where the desired field on `@document` is empty.
For example, here if ` @document.wxs_identifier.empty?` is true, the string `<<WXS_ID>>` will remain in the code snippet. This applies to all of the `unless` blocks below within this method as well.

app/components/earthworks/code_snippet_modal_component.rb
```diff
@@ -0,0 +1,115 @@
...
+
+      # replace WFS reference
+      unless @document.references.wfs.nil? || @document.references.wfs.endpoint.nil?
+        sub_content = sub_content.gsub('<<GEOSERVER_WFS>>', @document.references.wfs.endpoint)
+      end
+
+      # Replace Bounding box information
+      unless @document.geom_field.empty?
+        geometry = @document.geometry.geom
+        bbox_array = geometry.split('ENVELOPE(')[1].split(')')[0].split(',').map(&:strip)
```
marlo-longley, line 51: In some cases, documents have no geometry (although after a full reindex after https://github.com/sul-dlss/searchworks_traject_indexer/pull/1505 takes effect, I believe this case will no longer exist.) 
But for now, if a document has no geometry, I am getting this error when trying to parse the ENVELOPE:
```
undefined method `split' for nil

bbox_array = geometry.split('ENVELOPE(')[1].split(')')[0].split(',').map(&:strip)
```

app/javascript/controllers/codesnippet_controller.js
```diff
@@ -0,0 +1,40 @@
+import { Controller } from "@hotwired/stimulus"
+
+export default class extends Controller {
+    static targets = ["clipboardText", "clipboardCheck"];
+
+    async copyToClipboard(event) {  
+        let text = this.copyActivePaneContent();
```
jcoyne, line 7: Why not `const`?

app/javascript/controllers/codesnippet_controller.js
```diff
@@ -0,0 +1,40 @@
...
+        await navigator.clipboard.writeText(text);
+        this.displayCopyClick();
+    }
+
+    copyActivePaneContent() {
+      const codeText = document.querySelector("div#codeContent div.tab-pane.active pre code");
+      if(! codeText) {
+        console.log("Error copying content.  Code was not found");
```
jcoyne, line 15: should this be `console.error`?

app/javascript/controllers/codesnippet_controller.js
```diff
@@ -0,0 +1,40 @@
...
+    static targets = ["clipboardText", "clipboardCheck"];
+
+    async copyToClipboard(event) {  
+        let text = this.copyActivePaneContent();
+        await navigator.clipboard.writeText(text);
+        this.displayCopyClick();
+    }
+
+    copyActivePaneContent() {
+      const codeText = document.querySelector("div#codeContent div.tab-pane.active pre code");
```
jcoyne, line 13: can this be implemented at a stimulus target?  If not, can we use `this.element` rather than `document`?

app/components/earthworks/code_snippet_link_component.rb
```diff
@@ -0,0 +1,9 @@
+module Earthworks
+  # Display other options section in sidebar
```
marlo-longley, line 2: This comment seems copied from that component, possibly could confuse

app/views/catalog/code_snippet.html.erb
```diff
@@ -0,0 +1 @@
+<%= render Earthworks::CodeSnippetModalComponent.new(@docs.first) %>
```
marlo-longley, line 1: I think we can simplify this whole logic around `@docs` and the block you've added to the catalog controller by using the `document` that is passed via the sidebar controller ` Earthworks::OtherOptionsComponent.new(document: document)`

Happy to discuss this.

marlo-longley (changes requested):
Happy to discuss!

app/components/earthworks/other_options_component.rb
```diff
@@ -0,0 +1,11 @@
+module Earthworks
+  # Display other options section in sidebar
+  class OtherOptionsComponent < ViewComponent::Base
+    attr_reader :document
+
+    def initialize(document:)
+      @document = document
```
marlo-longley, line 7: I think this is redundant, given the `attr_reader` ?

app/views/catalog/code_snippet.html.erb
marlo-longley: We discussed this ~~ food for thought.

hudajkhan:
In addition to the changes above, I'll also look at a possible approach to handle comments in  #1093 either here or in a separate issue.

hudajkhan:
-- Remove code snippet for restricted datasets --

app/javascript/controllers/codesnippet_controller.js
hudajkhan: Copied from elsewhere. Will change it.

app/javascript/controllers/codesnippet_controller.js
hudajkhan: Will change.

app/components/earthworks/code_snippet_link_component.rb
hudajkhan: Will change to reflect what the component does

hudajkhan:
@dnoneill @marlo-longley @jcoyne I believe I addressed all your comments.  Thank you for your reviews!

app/components/earthworks/code_snippet_modal_component.html.erb
hudajkhan: I used your suggested approach and it worked fine. Thanks!

app/components/earthworks/code_snippet_modal_component.rb
hudajkhan: I tried to handle all the situations by replacing with a placeholder value where the document value was not present.

app/components/earthworks/code_snippet_modal_component.rb
hudajkhan: I tried to handle this by adding another check for @ document.geometry.geom.

app/javascript/controllers/codesnippet_controller.js
hudajkhan: I added a target for div#codeContent (which is the nav tab section) and then used the query selector directly.

app/components/earthworks/other_options_component.rb
hudajkhan: I removed this component entirely, based on discussions with @dnoneill .

hudajkhan:
@marlo-longley thanks!  I rebased and fixed the test (thanks for the suggestion @dnoneill )!

marlo-longley (approved):

PR #1132, opened 2024-08-07
move sidebar, update display

### LINKED ISSUE #646: Consider presenting Download and Export options differently
(I considered submitting this to the GeoBlacklight repo, since it seems like a common pattern other institutions might benefit from changing, but there is enough variation among institutions that I worry it might be too big of a change to impose in GeoBlacklight. @mejackreed will probably have a better sense of whether we do this locally or suggest for GeoBlacklight, if we decide to make a change.)

I'm concerned about the amount of sidebar space that is used by the download and export options, especially if we introduce some form of relations box in the sidebar. I'm also concerned about having a handful of similarly-looking boxes in the sidebar, some of which might be more important than others. (I'd like to rethink the Tools box as well, but will create a separate ticket for that.)

I also have some concerns about how the export and download affordances are presented. Those details combined with the use of boxes to contain everything make the sidebar kind of large and busy, in my opinion.

---

Screen Shot 2020-06-12 at 1 30 57 PM

---

### A - Downloads
- As best I can tell, items in Earthworks always, or almost always, only have one download available. Using a box to display one button feels unnecessary to me (although I get it was probably done for consistency).

- Again, there might be exceptions, but every download example I saw used a label that started with "Original." I wonder if the "Original" part is really necessary?

### B - Export formats
- This box often has several items in it, but I still wonder if placing the buttons in a box makes sense.

- The button labels are always the same: "Export" (probably not great for accessibility) and the export format is presented outside of the button. This feels necessary to me, and adds to the vertical space required (at least at some viewport widths).

### Proposed revision
Screen Shot 2020-06-12 at 2 00 04 PM

- Remove the Downloads and Export formats boxes

- Eliminate the use of "Original" in the download label

- For both download and export, use a single button to present the applicable options

- If there is more than one option, use a split button with the different download/export options

- I personally prefer the small size of the button (31px tall in my mockups versus 38px in the first screenshot of the current app), since these are secondary tasks and to reduce the vertical size of the sidebar, but if we use the split button approach this is less of an issue.

dnoneill:
closes #646 

Screenshot 2024-08-07 at 5 46 01 PM
Screenshot 2024-08-07 at 5 45 37 PM
Screenshot 2024-08-07 at 5 42 53 PM

Screenshot 2024-08-07 at 5 48 00 PM
Screenshot 2024-08-07 at 5 47 33 PM
Screenshot 2024-08-07 at 5 46 57 PM
Screenshot 2024-08-07 at 5 46 33 PM
Screenshot 2024-08-07 at 5 48 52 PM


Gemfile.lock
thatbudakguy, line 1: why the Gemfile.lock updates?

thatbudakguy (approved):

Gemfile.lock
dnoneill: Because uat would not let me deploy the branch to uat unless the dependencies were up to date and I couldn't find the skip check dependencies on this repo's deploy

PR #1149, opened 2024-08-09
Bl8 results component

### LINKED ISSUE #1143: Update search view to have icons on second line, no tooltips

dnoneill:
closes #1143 

Screenshot 2024-08-09 at 12 02 34 PM

### DIFF
--- a/app/assets/stylesheets/earthworks.css
+++ b/app/assets/stylesheets/earthworks.css
@@ -68,12 +68,3 @@
 .documents-list article:hover, .documents-list article:focus {
   box-shadow: -4px 4px 4px 0px var(--stanford-fog-light);
 }
-
-.document-counter {
-  display: none;
-}
-
-.documents-list .documentHeader .index_title {
-  margin-left: .75rem;
-  align-self: center;
-}
--- a/app/components/earthworks/header_icons_component.html.erb
+++ b/app/components/earthworks/header_icons_component.html.erb
@@ -1,0 +1,7 @@
+<div>
+  <% @fields.each do |field| %>
+    <% icon, label = get_icon(field) %>
+    <%= icon %>
+    <span class="me-1"><%= label %></span>
+  <% end %>
+</div>
--- a/app/components/earthworks/header_icons_component.rb
+++ b/app/components/earthworks/header_icons_component.rb
@@ -1,0 +1,18 @@
+# frozen_string_literal: true
+
+module Earthworks
+  class HeaderIconsComponent < Geoblacklight::HeaderIconsComponent
+    def get_icon(field)
+      icon_name = @document[field]
+      if icon_name&.include?('Datasets') && @document[Settings.FIELDS.RESOURCE_TYPE]
+        specific_icon = @document[Settings.FIELDS.RESOURCE_TYPE]
+        specific_icon = specific_icon.first if specific_icon.is_a?(Array)
+        specific_icon = specific_icon&.gsub(' data', '')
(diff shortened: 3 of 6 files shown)


app/assets/stylesheets/earthworks.css
```diff
@@ -69,11 +69,7 @@
...
-  display: none;
-}
-
-.documents-list .documentHeader .index_title {
-  margin-left: .75rem;
-  align-self: center;
+.documents-list .document {
+  padding-top: 0px;
+  margin-top: 0px;
 }
```
thatbudakguy, line 75: would it work to override or append to `classes` in the component to add `pt-0 mt-0` so we don't need to do it here?

app/components/geoblacklight/header_icons_component.rb
```diff
@@ -0,0 +1,27 @@
+# frozen_string_literal: true
+
+module Geoblacklight
+  class HeaderIconsComponent < ViewComponent::Base
```
thatbudakguy, line 4: could we just make this `Earthworks::HeaderIconsComponent` too? I think it's easier to know when things are changed if they get namespaced that way

dnoneill: Yeah I thought about that but then I think we need to pull out a bunch of components from geoblacklight and override those as well.

app/assets/stylesheets/earthworks.css
dnoneill: I can check but I am pretty sure I would need to pull out a blacklight component to override the classes.

dnoneill: Yup. I would need to override the partial https://github.com/projectblacklight/blacklight/blob/a8b5a3d3a2e8243c8a875d40528eae5c106d83bd/app/views/catalog/_document_list.html.erb#L3 to do this.

dnoneill: Actually I might be wrong. Nevermind ignore me. Pretty sure this is doable.

app/components/geoblacklight/header_icons_component.rb
thatbudakguy: unless there's more places that aren't in that search, I think that is OK (and in fact good). the things in that search are:
- the component itself, which we are already changing
- a partial, which we would change to use the new component
- a test, which is not in earthworks
- a helper function, which we would change to use the new component
- the component that references this component, which is already changed in this PR

that seems like a pretty easy lift to me! just two places that aren't already in the PR, and neither is a component.

dnoneill: I would be more inclined to change the way geoblacklight uses the component. Is there a reason we wouldn't want to update the config to set the headericon class in the controller?

thatbudakguy: that is another path, for sure. from what i've seen that is mostly used for "big" stuff like the document component, sidebar, etc. but if that seems easier to you, go for it

dnoneill: @thatbudakguy actually the helper I am pretty sure we don't want using our custom headericonscomponent. And the https://github.com/geoblacklight/geoblacklight/blob/9ce0dccdc2336dad520dfc0578743b20d421b0f6/app/views/catalog/_show_header_default.html.erb#L3 actually needs to get removed and fixed in geoblacklight (the document component missed this).

I am updating this PR.

thatbudakguy (approved):
lgtm, thanks!

PR #1150, opened 2024-08-09
fix show to match geoblacklight

dnoneill:


app/assets/stylesheets/earthworks.css
```diff
@@ -76,4 +76,8 @@
 .documents-list .documentHeader .index_title {
   margin-left: .75rem;
   align-self: center;
+}
+
+.help-text.viewer_protocol {
+  display: none;
 }
```
thatbudakguy, line 83: I know it isn't in the designs but I'm slightly worried about hiding this, if only because it makes it way easier in testing to see what viewer is actually getting rendered/what the data is supposed to be. Maybe we need to check if @dbranchini hid it intentionally, and if not if there's a way to re-add it to the page?

app/views/layouts/catalog_result.html.erb
```diff
@@ -6,6 +6,21 @@
...
     <%= yield %>
   </section>
+
+  <section class="col-lg-5">
+    <% if @document.file_format.include?('JPEG') && !@document[Settings.FIELDS[:GEOREFERENCED]] %>
+      <div role="alert" class="alert alert-info d-flex shadow-sm align-items-center">
+      <div class="bi bi-info-circle-fill fs-3 me-3 align-self-center
+                  d-flex justify-content-center"></div>
+      <div class="text-body">
+        <div>This map is not georeferenced.</div>
```
thatbudakguy, line 16: I know we don't do it everywhere, but imo it's good practice to put these strings in the locale file, at least for new content/templates

thatbudakguy (commented):
Could we split the georeference alert off of this? Would be easier to review just the display change, and then the added alert can come after

app/assets/stylesheets/earthworks.css
dnoneill: I will remove it for now and we can add it back in if Darcy says it was purposeful

app/views/layouts/catalog_result.html.erb
dnoneill: I removed the georeference alert. Making new branch.

thatbudakguy (approved):

app/assets/stylesheets/earthworks.css
dbranchini: I'm not sure what this is referring to? What did I eliminate? It probably wasn't intentional.

PR #1155, opened 2024-08-12
fix legacy icons, info_format

dnoneill:
See https://github.com/geoblacklight/geoblacklight/issues/1446. Since we don't have the fix implemented the icons aren't showing up. Adding the ICON_MAPPING fixes the icons. 
Screenshot 2024-08-12 at 1 38 54 PM

For the line `:INFO_FORMAT: "application/json"` see https://github.com/geoblacklight/geoblacklight/issues/612 and https://github.com/geoblacklight/geoblacklight/pull/1429. Switching to application/json allows highlighting on map when clicked.

### DIFF
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -225,7 +225,7 @@ WMS_PARAMS:
   :STYLES: ""
   :SRS: "EPSG:4326"
   :EXCEPTIONS: "application/json"
-  :INFO_FORMAT: "text/html"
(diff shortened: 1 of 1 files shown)


thatbudakguy:
could we get a little more context for this? a link to an issue or a more descriptive commit message would be helpful for posterity

dnoneill:
Sure. See updated comment.

PR #1161, opened 2024-08-12
Bl8 georeference alert

### LINKED ISSUE #1137: Show an alert if the item is not georeferenced
This should only display on items that could conceivably be georeferenced, but are not. One way to check that would be if we're going to render a IIIF viewer, since IIIF content is not georeferenced but might have a version that is.

The actual "is it georeferenced" value comes from [the `gbl_georeferenced_b` field](https://opengeometadata.org/ogm-aardvark/#georeferenced).

Screenshot 2024-08-08 at 09 54 31

dnoneill:
Screenshot 2024-08-12 at 4 32 11 PM

closes #1137


marlo-longley:
This looks good! I noticed that we don't have any fixture data with the field `gbl_georeferenced_b` set. Would it be worth adding this field to an appropriate fixture .json and adding a test?

marlo-longley (approved):

Issue #1138, opened 2024-08-08
Drop non-standard "theme" metadata values when indexing non-Stanford records


thatbudakguy:
We're moving towards "theme" as a replacement for "subject" in Earthworks because it's a small, controlled vocabulary and more useful to users than the great variety of existing subjects.

Unfortunately, although we use a translation map to ensure we don't add subjects that are out-of-vocabulary when indexing Stanford content, some OGM institutions have non-standard values in their data - record `berkeley-s7r13r` is an example.

We can add to the existing transformation block in our custom indexer to rewrite the values:
https://github.com/sul-dlss/earthworks/blob/d178bd870a8ab9bf0c3b9cd58099e2057be8fcd3/lib/earthworks/harvester.rb#L42-L51

For a list of valid values, see [the spec](https://opengeometadata.org/ogm-aardvark/#theme-values).

PR #1163, opened 2024-08-12
Filter dcat_theme_sm during OGM harvesting using controlled vocabulary

marlo-longley:
Closes https://github.com/sul-dlss/earthworks/issues/1138

I've not worked with the harvester before, so I welcome feedback if this is off base!


thatbudakguy (approved):
looks great!

Issue #1094, opened 2024-07-19
Colors for Index maps


dbranchini:
**Problem statement:**

I've seen a few variations on index maps. I'm not sure I fully understand the pattern of why one color scheme over another, but I did some research on this topic to see what colors would be color-blind safe. Of course, there are several different kinds of color blindness. I used this site to test a couple of variations -https://pilestone.com/pages/color-blindness-simulator-1, and I used these sites to find a color scheme that's color-blind safe - https://davidmathlogic.com/colorblind/#%23FFC20A-%230C7BDC and https://colorbrewer2.org/#type=diverging&scheme=PuOr&n=3. 

**Variations in the wild today:**

![Image](https://github.com/user-attachments/assets/510ff638-efd7-4531-a6af-3bf53aacd94a)

![Image](https://github.com/user-attachments/assets/729766aa-0b1a-47e8-95f3-fc98bb04ae70)

![Image](https://github.com/user-attachments/assets/6adc4d68-0dff-4d69-bec0-2ac9d1c12e44)

![Image](https://github.com/user-attachments/assets/81974aa4-9231-4908-90f0-32f0e05b4269)

**Proposed solution:**

For all index maps, use the following color palette: 

#d7191c - Maps we don't have
#ffffbf - Maps we have 
#2c7bb6 - Selected map (or the blue you're already using elsewhere)

Let's start with a 30-35% opacity too.

jcoyne:
#016895 is "sky dark" and #FFE781 is "Illuminating light" in the stanford scheme:
https://identity.stanford.edu/design-elements/color/accent-colors/

dbranchini:
I've reviewed this again, and we decided to go with reds and blues. I've updated the original description to avoid confusion. (Justin's colors will no longer be correct.) The [National Eye Institute](https://www.nei.nih.gov/learn-about-eye-health/eye-conditions-and-diseases/color-blindness/types-color-vision-deficiency) explains Red/Green and Blue/Yellow as the most common types of color blindness. Colorbrewer2.org allows you to select color-blind safe palettes. I decided on a Red/Blue palette for these maps and it's been tested on a color blindness simulator. 

Digital Red - #B1040E - Maps we don't have
Digital Blue - #006CB8 - Maps we do have

https://identity.stanford.edu/design-elements/color/web/

peetucket:
Do you have example object URLs (e.g. druids) that have the colors that are changing (that were used for those screenshots above)?

[...]

PR #1166, opened 2024-08-13
update index map colors

dnoneill:
part of #1094


thatbudakguy:
superseded by #1167?

dnoneill:
@thatbudakguy Nope. This one is going into main. When we do dependency updates on monday all our index maps should look the same and since I updated sul-embed, they are going to look different if we don't merge this.

thatbudakguy:
ah, I didn't notice the different branch target. thx

thatbudakguy (approved):

PR #1180, opened 2024-08-15
Smart suggest using search

### LINKED ISSUE #1047: Smart autocomplete
This is a brand-new feature to enhance the search experience, both from the homepage and search results page.

To try it out, visit the [interactive prototype](https://www.figma.com/proto/nkgg905s5FyXoyPZjVMco0/EarthWorks?page-id=0%3A1&node-id=190-4386&viewport=-3973%2C-836%2C0.58&t=5oPw70wWwebL9uPW-1&scaling=min-zoom&starting-point-node-id=190%3A4386&show-proto-sidebar=1) and click in the search bar to simulate typing.

The suggestions are grouped into categories:
- Datasets, Maps, etc. (corresponding to values in the [resource class](https://opengeometadata.org/ogm-aardvark/#resource-class) field in solr)
- Places

For the latter, the most basic implementation would be to use values from the ["Spatial coverage"](https://opengeometadata.org/ogm-aardvark/#spatial-coverage) field/facet. However, **this won't result in a spatial search** and won't move the map because those values are just metadata – they aren't tied to geospatial locations. We want to do more than this, if we can.

A more complete/featureful implementation would do [geocoding](https://en.wikipedia.org/wiki/Address_geocoding), so that "New york (city)" and "New york (state)" could be matched to geospatial coordinates. When you selected each option, you would go to a search with the spatial facet active, as though you had already moved the map to the coordinates/bbox of the place you selected.

This deserves some thought re: implementation. We could query some kind of web service live during the autocomplete process, or we could pre-compute a list of known locations and their coordinates to make the process faster/simpler (but perhaps less complete).

Note that Stanford maintains its own geocoder service at https://sul-geocoding-web.stanford.edu/. It would be great to take advantage of this! There's more information on that page about how the service works; see in particular [this page on the geocoding API](https://locator.stanford.edu/arcgis/sdk/rest/index.html#/Geocode_Addresses/02ss00000040000000/). There is also a [suggestion service](https://locator.stanford.edu/arcgis/sdk/rest/index.html#/Suggest/02ss000000q2000000/) that might be worth investigating.

Whatever we do, we probably want to take advantage of solr's [built-in functionality for search suggestions](https://solr.apache.org/guide/solr/latest/query-guide/suggester.html).

edsu:
Implement "smart suggest" using the Solr select endpoint rather than suggest in order to get the document IDs for the title results, so they can be directly linked to.

When the user types in a string a search for the string with `*` appended of `dct_title_ti` and `dct_spatial_tmi` is performed. The results are stored in an instance of the new `Suggestions` class which has methods to return locations, datasets and maps. Each of the suggestions is an instance of `Suggestion` which can highlight the query.

TODO: the Suggester, Suggestions and Suggestion classes need unit tests. I tested with 4k documents and it seemed to perform similarly to the suggest endpoint, but it will be useful to see how it performs with more.

Closes #1047

### DIFF
--- a/app/controllers/catalog_controller.rb
+++ b/app/controllers/catalog_controller.rb
@@ -3,6 +3,7 @@ require 'blacklight/catalog'
 class CatalogController < ApplicationController
   include BlacklightRangeLimit::ControllerOverride
   include Blacklight::Catalog
+  include Blacklight::Searchable
 
   configure_blacklight do |config|
     # Ensures that JSON representations of Solr Documents can be retrieved using
@@ -340,5 +341,10 @@ class CatalogController < ApplicationController
         # Otherwise draw the full page
       end
     end
+  end
+
+  def suggest
+    @suggestions = Suggester.suggest(search_service, params[:q])
+    render layout: false
   end
 end
--- a/app/models/suggester.rb
+++ b/app/models/suggester.rb
@@ -1,0 +1,110 @@
+class Suggester
+  def self.suggest(search_service, query)
+    new(search_service, query).suggestions
+  end
+
+  def initialize(search_service, query, max_results: 50)
+    @query = query
+    @search_service = search_service
+    @max_results = max_results
+  end
+
+  def suggestions
+    make_suggestions(search)
+  end
+
+  private
+
+  def search
+    solr_q = "dct_title_ti:#{@query}* OR dct_spatial_tmi:#{@query}*"
+    @search_service.repository.search(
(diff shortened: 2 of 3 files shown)


edsu:
We decided not to pursue using a Search approach. This was mostly because using the newly configured suggest index in #1154 will guarantee that a certain number of matches from titles and locations are returned, which is not the case when doing a single search against the title and location fields.

PR #1181, opened 2024-08-15
Use container-fluid and set max-width to 1600px

### LINKED ISSUE #1177: Allow the main container to expand further
Our current designs for desktop were done with a width of 1608px, but we're currently using bootstrap's default `container` class for the main content, which is often substantially smaller than that on most screens. The result is that the UI looks a little cramped.

To allow more breathing room, especially on the show page, let's try changing the `.container` to a `.container-fluid` and then set a max-width of around 1600px so that it doesn't become too wide on very large screens.

marlo-longley:
Closes #1177

Looks like:
Screenshot 2024-08-15 at 13 16 25
Screenshot 2024-08-15 at 13 16 38

### DIFF
--- a/app/assets/stylesheets/earthworks.css
+++ b/app/assets/stylesheets/earthworks.css
@@ -2,6 +2,12 @@
   --stanford-50-black: #979694;
 }
 
+/* Selects both #main-content and navbar containers */
+#main-container, .navbar-search > * {
+  max-width:  1600px;
+  padding: 0 60px;
+}
+
 .header-icons {
   color: var(--stanford-50-black);
   font-size: .875rem;
--- a/app/controllers/catalog_controller.rb
+++ b/app/controllers/catalog_controller.rb
@@ -25,6 +25,9 @@ class CatalogController < ApplicationController
       q: "{!raw f=#{Settings.FIELDS.ID} v=$id}"
     }
 
+    # When set to true, Blacklight uses container-fluid as the layout container
+    config.full_width_layout = true
+
     # GeoBlacklight Defaults
     # * Adds the "map" split view for catalog#index
     config.view.split(partials: ['index'])
--- a/app/views/catalog/_home_text.html.erb
+++ b/app/views/catalog/_home_text.html.erb
@@ -1,4 +1,4 @@
-<div class='container mt-4 mb-4'>
+<div class='mt-4 mb-4'>
   <div class='row'>
     <div class='col-sm'>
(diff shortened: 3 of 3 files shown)


marlo-longley:
@dbranchini I would love a visual check on this PR!

app/assets/stylesheets/earthworks.css
```diff
@@ -2,6 +2,12 @@
   --stanford-50-black: #979694;
 }
 
+/* Selects both #main-content and navbar containers */
+ .container-fluid {
+  max-width:  1600px;
```
jcoyne, line 7: This seems like you're going to confuse people. A fluid container is defined as not having any max width and no padding.  Are you sure you don't want to add `.container-xxl` instead?

marlo-longley: Thanks for reviewing Justin. I was not part of any discussions about the ticket parameters. I just know that the design called for a breakpoint at 1600 and 60px of padding.

If adding this CSS to the `container-fluid` class feels like it's going against good practice, maybe we could add an additional class and style that.

jcoyne: Something like this: https://github.com/sul-dlss/earthworks/pull/1183

marlo-longley: Thanks. This looks good but potentially does not align with the designer spec -- I am not sure how strict that is. Let me check.

marlo-longley: Hi @jcoyne - appreciate that PR example. I just chatted with the workcycle team and learned that we actually do want the behavior of `container-fluid` at all sizes, so even at those smaller breakpoints we want to take up 100% minus 60px.

So I am going to try a different approach than those breakpoints and update this PR.

marlo-longley: OK, basically it's the same approach, just not targeting `.container-fluid` ...

app/assets/stylesheets/earthworks.css
```diff
@@ -2,6 +2,12 @@
   --stanford-50-black: #979694;
 }
 
+/* Selects both #main-content and navbar containers */
+#main-container, .navbar-search > * {
+  max-width:  1600px;
+  padding: 0 60px;
```
jcoyne, line 8: could you set the variable instead?
```css
--bs-gutter-width-x: 120px
```

dnoneill (approved):

PR #1184, opened 2024-08-15
Override Blacklight's StartOverButtonComponent

### LINKED ISSUE #1176: get start over button to not redirect to the homepage
In order to do this it should be pretty simple. You should be able to override the call method here: https://github.com/projectblacklight/blacklight/blob/a8b5a3d3a2e8243c8a875d40528eae5c106d83bd/app/components/blacklight/start_over_button_component.rb#L5-L7 and add the params `search_field=all_fields`

marlo-longley:
Closes #1176

### DIFF
--- a/app/components/blacklight/start_over_button_component.rb
+++ b/app/components/blacklight/start_over_button_component.rb
@@ -1,0 +1,24 @@
+# frozen_string_literal: true
+
+# Overrides the default Blacklight StartOverButtonComponent in order to
+# always set the search_field to 'all_fields'.
+module Blacklight
+  class StartOverButtonComponent < Blacklight::Component
+    def call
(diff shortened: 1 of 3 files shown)


dnoneill:
You should be able to create an Earthworks StartButton that you can override with this in the catalog_controller https://github.com/projectblacklight/blacklight/blob/a8b5a3d3a2e8243c8a875d40528eae5c106d83bd/app/components/blacklight/constraints_component.rb#L28

dnoneill:
My bad. Thought it was somewhere else. Wasn't paying close enough attention.

marlo-longley:
We decided to try and get an upstream Blacklight PR for this instead. Drafting this one until then.

PR #1190, opened 2024-08-16
update color for secondary buttons

dnoneill:


app/assets/stylesheets/earthworks.css
```diff
@@ -2,6 +2,16 @@
   --stanford-50-black: #979694;
 }
 
+
+.btn-outline-secondary {
```
jcoyne, line 6: I suspect it's going to get confusing that we're using blue as the secondary color, since the component library defines blue as the primary color.   Is it possible to change the classes on the buttons?

dnoneill: Not without changing backlight. Specifically it is these buttons: https://github.com/projectblacklight/blacklight/blob/a8b5a3d3a2e8243c8a875d40528eae5c106d83bd/app/helpers/blacklight/component_helper_behavior.rb#L26C1-L30C8  so maybe there is a way to override this class? I need to check to see what actions can do.

jcoyne: I think that's why I added https://github.com/projectblacklight/blacklight/pull/3267

dnoneill: makes sense. Will leave as is for now and wait until this gets merged.

PR #1191, opened 2024-08-16
add expand/collapse buttons to search results

### LINKED ISSUE #1185: Expand/compact view

dnoneill:
Screenshot 2024-08-16 at 2 43 30 PM

closes #1185

### DIFF
--- a/app/assets/stylesheets/earthworks.css
+++ b/app/assets/stylesheets/earthworks.css
@@ -152,3 +152,10 @@ Uncomment lines 142-144
     display: none;
   }
 }
+
+.expand-collapse .btn {
+  --bs-btn-disabled-color: var(--bs-btn-bg);
+  --bs-btn-disabled-bg: var(--bs-btn-color);
+  --bs-btn-disabled-border-color:var(--bs-btn-border-color);
+  --bs-btn-disabled-opacity: 1;
+}
--- a/app/javascript/controllers/expandcollapse_controller.js
+++ b/app/javascript/controllers/expandcollapse_controller.js
@@ -1,0 +1,16 @@
+import { Controller } from "@hotwired/stimulus"
+
+export default class extends Controller {
+  toggleDescriptions(evt) {
+    const activeClass = 'active'
+    const currentActive = this.element.querySelector(`.${activeClass}`)
+    currentActive.classList.remove(activeClass)
+    currentActive.disabled = false;
+    evt.currentTarget.disabled = true;
+    evt.currentTarget.classList.add(activeClass)
+    const elements = Array.from(document.getElementsByClassName('more-info-area'));
+    elements.forEach(function(element) {
+      element.children[0].classList.toggle("collapse");
+    });
+  }
+}
--- a/app/views/catalog/_sort_and_per_page.html.erb
+++ b/app/views/catalog/_sort_and_per_page.html.erb
@@ -1,0 +1,14 @@
+<div id="sortAndPerPage" class="sort-pagination d-md-flex justify-content-between" role="navigation" aria-label="<%= t('blacklight.search.per_page.aria_label')%>">
+  <%= render partial: "paginate_compact", object: @response if show_pagination? %>
+  <div class="search-widgets">
+    <%= render_results_collection_tools %>
+    <div class="expand-collapse btn-group" data-controller="expandcollapse">
+      <button class="btn btn-outline-secondary active" aria-label="collapse descriptions" disabled data-action="expandcollapse#toggleDescriptions">
+        <i class="bi bi-list-ul"></i>
+      </button>
+      <button class="btn btn-outline-secondary" aria-label="expand descriptions" data-action="expandcollapse#toggleDescriptions">
+        <i class="bi bi-text-paragraph"></i>
+      </button>
+    </div>
+  </div>
+</div>


app/assets/stylesheets/earthworks.css
```diff
@@ -152,3 +152,10 @@ Uncomment lines 142-144
     display: none;
   }
 }
+
+.expand-collapse .btn:disabled {
```
jcoyne, line 156: I don't think you need the `:disabled` pseudo selector here.

dnoneill: I am not sure I understand your meaning. I want these styles to show up only when the button is disabled.

jcoyne: They will. You set variables for disabled styles.

app/views/catalog/_sort_and_per_page.html.erb
```diff
@@ -0,0 +1,10 @@
+<div id="sortAndPerPage" class="sort-pagination d-md-flex justify-content-between" role="navigation" aria-label="<%= t('blacklight.search.per_page.aria_label')%>">
+  <%= render partial: "paginate_compact", object: @response if show_pagination? %>
```
jcoyne, line 2: if you use "partial", don't you also need to use "locals". I prefer to omit partial.

jcoyne (changes requested):

app/javascript/controllers/expandcollapse_controller.js
```diff
@@ -0,0 +1,17 @@
...
+export default class extends Controller {
+  toggleDescriptions(evt) {
+    const activeClass = 'active'
+    const currentActive = this.element.querySelector(`.${activeClass}`)
+    currentActive.classList.remove(activeClass)
+    currentActive.disabled = false;
+    const target = evt.target.classList.contains('bi') ? evt.target.parentElement : evt.target;
+    target.disabled = true;
+    target.classList.add(activeClass)
+    const elements = Array.from(document.getElementsByClassName('more-info-area'));
```
jcoyne, line 12: This is a pretty fragile way to bind to these elements.  Have you considered making them stimulus targets for this controller?  Alternatively, you could use stimulus outlets to reference another controller responsible for each document. https://stimulus.hotwired.dev/reference/outlets or even just emit an custom event like "showList"/"showParagraph", that another controller will bind to.

app/views/catalog/_sort_and_per_page.html.erb
```diff
@@ -0,0 +1,10 @@
...
+  <div class="search-widgets">
+    <%= render_results_collection_tools %>
+    <div class="expand-collapse btn-group" data-controller="expandcollapse">
+      <button class="btn btn-outline-secondary active" disabled data-action="expandcollapse#toggleDescriptions"><i class="bi bi-list-ul"></i></button>
```
jcoyne, line 6: These buttons don't look accessible to me.

jcoyne (changes requested):

app/views/catalog/_sort_and_per_page.html.erb
dnoneill: This is copied directly from blacklight.

app/assets/stylesheets/earthworks.css
dnoneill: Gotcha. That makes sense.

app/javascript/controllers/expandcollapse_controller.js
dnoneill: I thought about the stimulus targets but then you have to figure out which got clicked. I will think about it some more maybe.

app/views/catalog/_sort_and_per_page.html.erb
jcoyne: I was wrong. This is actually important.

app/javascript/controllers/expandcollapse_controller.js
jcoyne: I still think that binding behaviors to css styles (`.more-info-area`) is not very robust.  I urge you to look at the solutions that are built into Stimulus for this.

app/views/catalog/_sort_and_per_page.html.erb
```diff
@@ -0,0 +1,14 @@
+<div id="sortAndPerPage" class="sort-pagination d-md-flex justify-content-between" role="navigation" aria-label="<%= t('blacklight.search.per_page.aria_label')%>">
+  <%= render partial: "paginate_compact", object: @response if show_pagination? %>
+  <div class="search-widgets">
+    <%= render_results_collection_tools %>
+    <div class="expand-collapse btn-group" data-controller="expandcollapse">
+      <button class="btn btn-outline-secondary active" aria-label="collapse descriptions" disabled data-action="expandcollapse#toggleDescriptions">
```
jcoyne, line 6: Shouldn't these also have `aria-controls`?  If one of them always has to be selected, can this just be a styled radio button (e.g. hide the button entirely and display a image in the label)?

app/javascript/controllers/expandcollapse_controller.js
dnoneill: Gotcha I misunderstood the area the comment was pointing towards, for some reason my brain said the top line and not the last line. Makes sense.

app/views/catalog/_sort_and_per_page.html.erb
dnoneill: Please see https://github.com/sul-dlss/earthworks/pull/1234/files to see if this is what you meant for both this and the previous comment. Also if you know the css off the top of your head for styling this I would appreciate it. What is in there now doesn't work.

PR #1194, opened 2024-08-16
update overlays on maps, update importmap to use latest JS from geoblacklight

dnoneill:


hudajkhan:
With changes, the overlay map color on the search page now looks like

Screenshot 2024-08-27 at 10 02 05 AM


As compared to UAT:
Screenshot 2024-08-27 at 10 05 13 AM

hudajkhan:
I was still seeing some blue boxes for some other pages and informed @dnoneill.  Appears a solution will be forthcoming.

marlo-longley:
Not to be too picky, but it might be helpful for future-us if this PR or commit wording referenced the fact that we are getting in line with the big upstream GBL change. Because if I'm reading this correctly the changes to `settings.yml` are related to "update overlays on maps" but in general this is the PR that is upgrading Earthworks to the new setup by removing reference to all of GBL's external dependencies.

dnoneill:
@marlo-longley How do you feel about updating the commit message to reflect the PR instead?

marlo-longley:
I just want us to be able to track down the commit where we changed how we import GBL JavaScript.

hudajkhan:
Individual page with bounding box view (not using embed viewer) is working as expected now.

hudajkhan (approved):

PR #1215, opened 2024-08-20
update header/footer margin to match main container

### LINKED ISSUE #1208: align header and footer margins with content

dnoneill:
closes #1208

### DIFF
--- a/app/assets/stylesheets/earthworks.css
+++ b/app/assets/stylesheets/earthworks.css
@@ -3,7 +3,7 @@
 }
 
 /* Selects both #main-content and navbar containers */
-#main-container, .navbar-search > * {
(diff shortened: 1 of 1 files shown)


app/assets/stylesheets/earthworks.css
```diff
@@ -3,7 +3,7 @@
 }
 
 /* Selects both #main-content and navbar containers */
-#main-container, .navbar-search > * {
+#main-container, .navbar-search > *, .stanford-header-container, .stanford-footer-container {
```
thatbudakguy, line 6: instead of adding these, could we just use the IDs/classes of the things that contain the container? so:
- `.masthead .container`
- `#pre-footer .container`
- `#su-footer .container`

I think those are easier to read and tell what they refer to.

thatbudakguy (approved):
thanks!

PR #1221, opened 2024-08-20
update selections start over button and order

### LINKED ISSUE #1209: update selected facets to match figma
Screenshot 2024-08-19 at 4 10 09 PM

dnoneill:
closes #1209


app/assets/stylesheets/earthworks.css
```diff
@@ -141,4 +141,38 @@ Uncomment lines 142-144
...
 /* .search-query-form .search-btn .blacklight-icons-search {
   display: none;
-} */
\ No newline at end of file
+} */
+
+.filter .btn-outline-secondary {
+  --bs-btn-border-color: var(--stanford-fog-light);
+  --bs-btn-color: var(--stanford-black);
+  background-color: var(--stanford-fog-light);
```
jcoyne, line 159: Why do you need this?  Can you set `--bs-btn-bg` instead?

app/assets/stylesheets/earthworks.css
```diff
@@ -141,4 +141,38 @@ Uncomment lines 142-144
...
+  --bs-btn-bg-color: var(--stanford-fog-light);
```
jcoyne, line 160: is this a typo? should it be `--bs-btn-bg`?

app/assets/stylesheets/earthworks.css
```diff
@@ -152,3 +152,37 @@ Uncomment lines 142-144
...
+  --bs-btn-active-border-color: var(--stanford-fog-light);
+  --bs-btn-active-color: var(--stanford-black);
+  --bs-btn-padding-y: .25rem;
+}
+
+.applied-filter .remove:hover {
+  --bs-btn-hover-border-color: var(--stanford-fog-light);
+  --bs-btn-hover-bg: var(--stanford-fog-light);
+  --bs-btn-hover-color: var(--stanford-digital-red);
+  text-decoration: none;
```
jcoyne, line 172: Can you use `--bs-link-hover-decoration` instead?

jcoyne (changes requested):

app/assets/stylesheets/earthworks.css
```diff
@@ -152,3 +152,37 @@ Uncomment lines 142-144
...
+.filter .constraint-value {
+  border-radius: 99px 0px 0px 99px;
+  padding-right: .25rem;
+}
+
+.filter .remove {
+  border-radius: 0px 99px 99px 0px;
+}
+
+.constraint-value, .remove {
```
jcoyne, line 184: Can you explain why we need this rule.  I don't see any difference before and after applying this.

app/assets/stylesheets/earthworks.css
jcoyne: If you use the variables, you don't need the `:hover` pseudo-selector.

app/assets/stylesheets/earthworks.css
```diff
@@ -152,3 +152,37 @@ Uncomment lines 142-144
...
+  --bs-btn-hover-color: var(--stanford-digital-red);
+  text-decoration: none;
+}
+
+.filter .constraint-value {
+  border-radius: 99px 0px 0px 99px;
+  padding-right: .25rem;
+}
+
+.filter .remove {
```
jcoyne, line 180: Aren't `.filter` and `.applied-filter` on the same elements?  Why is this one separate?

app/assets/stylesheets/earthworks.css
dnoneill: I isn't centered when you don't apply this. The text is slightly closer to the top then the bottom.

app/assets/stylesheets/earthworks.css
dnoneill: Nope. Unfortunately geoblacklight css has the line 

a {
  color: rgba(var(--bs-link-color-rgb), var(--bs-link-opacity, 1));
  text-decoration: underline;
}

app/assets/stylesheets/earthworks.css
dnoneill: yup

app/assets/stylesheets/earthworks.css
dnoneill: I could swear that didn't work when I tried it, but it does. Now changed.

app/assets/stylesheets/earthworks.css
jcoyne: Can we change that in Geoblacklight? Is seems like it's an impediment for us, so it will likely cause others trouble too.  When you run into something like this, I find it helpful if you write in a comment why you are doing it.

app/assets/stylesheets/earthworks.css
dnoneill: good catch. They should all be applied-filter. I noticed later in the process that filter isn't applied to queries, but applied-filter is. I have updated all the css to that.

app/assets/stylesheets/earthworks.css
jcoyne: Can you show me where you see that? I'm not seeing it on https://earthworks-uat.stanford.edu/?f%5Bdct_publisher_sm%5D%5B%5D=Collins+Bartholomew+Ltd.&q=&search_field=all_fields
Screenshot 2024-08-20 at 4 48 05 PM

jcoyne: Are you sure it's not https://github.com/sul-dlss/component-library/blame/main/styles/button.css#L2 that you're hitting.  I'm pretty sure DLUX said that the hover state of buttons should have underline.

dnoneill: Nope. Element above.

dnoneill: @jcoyne 
Screenshot 2024-08-20 at 5 59 26 PM

jcoyne: Let's work with the designers to understand what they want for the component library, rather than having to override it in every project where we use the component library.

app/components/earthworks/constraints_component.rb
```diff
@@ -0,0 +1,4 @@
+module Earthworks
+  class ConstraintsComponent < Blacklight::ConstraintsComponent
```
jcoyne, line 2: Can you add documentation about why this override is required.

app/assets/stylesheets/earthworks.css
dnoneill: Should I remove the line for now?

dnoneill: Sounds good.

app/assets/stylesheets/earthworks.css
jcoyne: Can this be pushed up to Blacklight?  It seems to me this affects every Blacklight site, right?

jcoyne: Actually, it doesn't affect demo.projectblacklight.org, so where is this difference coming from?

marlo-longley:
Update from design team: this should go into the component library

app/components/earthworks/constraints_component.rb
```diff
@@ -0,0 +1,9 @@
+# Override allows for:
+# 1. moving of start button to end of selected items
+# 2. constraints-label not being visually-hidden
+# 3. To override default query_constraint_component, facet_constraint_component
+# so we can get the HTML in for creating pills from the component library
+module Earthworks
```
jcoyne, line 6: Can you remove the Earthworks module from this component?

jcoyne (changes requested):

app/components/earthworks/constraints_component.rb
dnoneill: done

app/components/constraints_component.rb
```diff
@@ -0,0 +1,10 @@
+# Override allows for:
+# 1. moving of start button to end of selected items
```
jcoyne, line 2: 👍

app/components/start_over_button_component.rb
```diff
@@ -0,0 +1,22 @@
+# frozen_string_literal: true
+
+# Overrides the default Blacklight StartOverButtonComponent in order to
+# always set the search_field to 'all_fields' and update btn class to btn-outline-primary
```
jcoyne, line 4: 👍

jcoyne (approved):

PR #1222, opened 2024-08-20
add other icon

### LINKED ISSUE #1218: Other icon under resource class
I suggest a question mark (although seems more appropriate for unknown) or a simple but thick dash for the other icon under resource class.

dnoneill:
closes #1218

### DIFF
--- a/app/components/blacklight/icons/other_component.rb
+++ b/app/components/blacklight/icons/other_component.rb
@@ -1,0 +1,23 @@
+# frozen_string_literal: true
+
+module Blacklight
+  module Icons
+    class OtherComponent < Blacklight::Icons::IconComponent
+      self.svg = svg
+
+      def svg
+        <<~SVG
+          <svg xmlns="http://www.w3.org/2000/svg" width="16" height="16" fill="currentColor" class="bi bi-question-circle" viewBox="0 0 16 16">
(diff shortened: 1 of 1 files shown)


app/components/blacklight/icons/other_component.rb
```diff
@@ -0,0 +1,23 @@
+# frozen_string_literal: true
+
+module Blacklight
+  module Icons
+    class OtherComponent < Blacklight::Icons::IconComponent
```
thatbudakguy, line 5:
```suggestion
    class OtherResourceClassComponent < Blacklight::Icons::IconComponent
```
less confusing?

dnoneill: We will have to add a mapping into the config if we do this.

thatbudakguy: I guess it's already in the `Icons` namespace, so not as bad as it could be...

thatbudakguy (approved):

app/components/blacklight/icons/other_component.rb
dnoneill: We can do it, we would have to add the line 
other-resource: other here:
https://github.com/sul-dlss/earthworks/blob/bl8/config/settings.yml#L465-L466

PR #1224, opened 2024-08-20
Also available links component/all ref links in more details

### LINKED ISSUE #1151: DOI link display
~~Blocked by #1114~~

Once we have DOI links in the index, think about how to display them.
The DOI are indexed in an array with the purl in `'dct_identifier_sm'`
Currently the purl displays as part of "More details at" -- do we want add the DOI here?

(see bottom of image)
Screenshot 2024-08-08 at 16 22 18

peetucket:
Fixes #1151 

- Remove searchworks URL in sidebar as it's own display
- Add new component that shows searchworks URL (if available)
- Show any links in identifiers in the "More details at" section in the main metadata display (which could be an array)

See examples below for the extra links and new searchworks display:

Screenshot 2024-08-21 at 11 00 27 AM
Screenshot 2024-08-21 at 11 00 36 AM


app/helpers/earthworks_blacklight_helper.rb
```diff
@@ -7,4 +7,10 @@ def show_content_classes
   def show_sidebar_classes
     'page-sidebar col-lg-2'
   end
+
+  def render_details_links(args)
+    args[:document].identifiers.map do |identifier|
+      link_to identifier, identifier
+    end.join(', ').html_safe
```
peetucket, line 14: rubocop doesn't like `.html_safe` -- is there a better way of doing this via some blacklight magic (i.e. `add_show_field` can render links)

app/components/earthworks/also_available_component.html.erb
```diff
@@ -0,0 +1,8 @@
+<div class="h4">Also available at</div>
```
dnoneill, line 1: This needs to be a h2

peetucket: ok

app/controllers/catalog_controller.rb
```diff
@@ -318,11 +318,6 @@ class CatalogController < ApplicationController
                                                      (Settings.METADATA_SHOWN &
                                                       options[:document].references.refs.map { |x| x.type.to_s }).any?
                                                  }
-    config.add_show_tools_partial :searchworks_url, component: Earthworks::SearchworksUrl,
-                                                    if: proc { |_context, _config, options|
-                                                          options[:document] &&
-                                                            options[:document].searchworks_url.present?
-                                                        }
```
peetucket, line 325: this is now replaced by the new component for the sidebar, which is also more flexible in that it allows additional links to be added in later

app/components/earthworks/also_available_component.html.erb
dnoneill: Sorry I wasn't clear. Should be 
<h2 class="h4">

peetucket: fixed

marlo-longley:
This markup breaks the layout flow

Screenshot 2024-08-22 at 12 33 04

marlo-longley (changes requested):
This markup breaks the layout flow

Screenshot 2024-08-22 at 12 33 04

app/components/earthworks/also_available_component.html.erb
```diff
@@ -0,0 +1,8 @@
+<h2 class="h4">Also available at</div>
```
marlo-longley, line 1: @peetucket this is what is breaking the layout. It should read:
`<h2>Also available at</h2>`
It shouldn't close the wrapping `</div>`

marlo-longley:
Also, now that the CodeSnippetComponent is added to the show tools, we need to rebase this PR.

app/components/earthworks/also_available_component.html.erb
peetucket: thanks - fixed closing tag (but left `class="h4"`, which I think it still needed?)

app/models/solr_document.rb
```diff
@@ -28,6 +28,11 @@ def mixed?
     fetch(Settings.FIELDS.RESOURCE_CLASS, []) == 'Other'
   end
 
+  # a list of all the available other urls for the document
+  def also_available_links
```
marlo-longley, line 32: I almost feel like this method should go into the component. It is relying on `searchworks_url` which makes sense as a method on the solr_document since it's pulling from a solr field. But this is an extra layer on top of that -- it's not directly related to the solr_document. Thoughts?

marlo-longley:
One final question then I think I'm good --
Was it intentional that this section moved above Other options as opposed to the old order?

Screenshot 2024-08-22 at 14 56 24
Screenshot 2024-08-22 at 14 56 32

app/models/solr_document.rb
peetucket: makes sense -- i'll move to the component

peetucket:
> One final question then I think I'm good -- Was it intentional that this section moved above Other options as opposed to the old order?
> Screenshot 2024-08-22 at 14 56 24 Screenshot 2024-08-22 at 14 56 32

This new component is rendered at the bottom of the new sidebar component: https://github.com/sul-dlss/earthworks/pull/1224/files#diff-4249e29ed8d4a3d20d40fe550149bc13a63556164699bd0d4af329fa1a342428R12  The previous Searchworks link only panel was rendered differently, hence the different order: https://github.com/sul-dlss/earthworks/blob/main/app/controllers/catalog_controller.rb#L316-L322

app/components/earthworks/also_available_component.html.erb
```diff
@@ -0,0 +1,8 @@
+<h2 class="h4">Also available at</h2>
+<div>
+    <ul class="list-unstyled">
+        <% document.also_available_links.each do |label, url| %>
```
marlo-longley, line 4: do you need to change the call now?

peetucket: indeed, good catch

app/components/earthworks/also_available_component.rb
```diff
@@ -0,0 +1,22 @@
+# frozen_string_literal: true
+
+module Earthworks
+  # Display also available as links in sidebar
+  class AlsoAvailableComponent < ViewComponent::Base
+    attr_reader :document
+
+    def initialize(document:, **)
```
jcoyne, line 8: do we need the **?

jcoyne (changes requested):

app/components/earthworks/also_available_component.rb
peetucket: nope, not sure why i put that there (probably copy-pasta)

app/components/earthworks/also_available_component.html.erb
```diff
@@ -0,0 +1,8 @@
+<h2 class="h4">Also available at</h2>
+<div>
```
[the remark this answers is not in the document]
peetucket: i guess not - no styling on it, i'll remove

app/controllers/catalog_controller.rb
```diff
@@ -323,11 +323,7 @@ class CatalogController < ApplicationController
                                                         options[:document] &&
                                                           !options[:document].restricted?
                                                       }
-    config.add_show_tools_partial :searchworks_url, component: Earthworks::SearchworksUrl,
```
jcoyne, line 326: Can we remove the SearchworksUrl now? I don't think anything else calls it.

jcoyne (changes requested):

app/controllers/catalog_controller.rb
peetucket: Yes, that component was removed as part of this PR.

marlo-longley (approved):

app/controllers/catalog_controller.rb
marlo-longley: Oops. @jcoyne let us know if this looks OK to you!

jcoyne (approved):

PR #1227, opened 2024-08-20
allow for file_format to be nil

dnoneill:


app/views/layouts/catalog_result.html.erb
```diff
@@ -8,7 +8,7 @@
   </section>
 
   <section class="col-lg-5">
-    <% if (@document.file_format.include?('JPEG') || @document.item_viewer.iiif) && !@document[Settings.FIELDS[:GEOREFERENCED]] %>
+    <% if (@document.file_format&.include?('JPEG') || @document.item_viewer.iiif) && !@document[Settings.FIELDS[:GEOREFERENCED]] %>
```
jcoyne, line 11: Can you make this a `jpeg?` method in SolrDocument?

jcoyne (changes requested):

app/models/solr_document.rb
```diff
@@ -50,4 +50,10 @@ def searchworks_url
   # and Blacklight::Solr::Document#to_semantic_values
   # Recommendation: Use field names from Dublin Core
   use_extension(Blacklight::Document::DublinCore)
+
+  # Check to see if view is an image
+  # Used for georeferencing message
+  def is_image?
+    file_format&.include?('JPEG')
```
jcoyne, line 57: Would you please add a test for this method, including a case when file_format is nil?

jcoyne (changes requested):

jcoyne (approved):
Thanks!

PR #1229, opened 2024-08-21
move tools to header

### LINKED ISSUE #673: Move show tools (bookmark, cite, email) to show page header
See https://github.com/projectblacklight/blacklight/issues/2316 for the recommendation to move the core Blacklight tools (bookmark, cite, email, etc.) out of the sidebar. If this is done, the Earthworks sidebar can be focused on the geo-specific tools in one sidebar box, and use another for the Relations stuff, etc.

dnoneill:
closes #673 

Screenshot 2024-08-23 at 3 34 59 PM
Screenshot 2024-08-23 at 3 34 49 PM

@dbranchini This can't go into production until backlight is released but if you want you can review the screenshots attached to this PR. Also there are two differences I can see from the Figma that I didn't address for the following reasons.
1. The line goes past the tools, didn't fix because it was going to be a pain
2. The tools in the upper right are underlined. In order to do this you have to underline the icons as well which I think looks weird. 
Screenshot 2024-08-23 at 3 43 42 PM


app/assets/stylesheets/geoblacklight.scss
```diff
@@ -1,3 +1,3 @@
 @import 'bootstrap/scss/bootstrap';
-@import 'blacklight-frontend/app/assets/stylesheets/blacklight/blacklight';
+@import 'blacklight/blacklight';
```
peetucket, line 2: is it ok to commit this change and the `bin/build-css` file? (not questioning it so much as verifying it's ok)

thatbudakguy (approved):

dbranchini:
The line going past the tools looks fine. I guess what bothers me a little is that the line aligns with the map on the right, but it doesn't align with the sidebar on the left. 

The lack of underline should be fine. In cases where the link is within a body of text, it should be accompanied by an indicator (in addition to color) to signify that it's a link. I don't think it applies here.

dnoneill:
I am going to open a separate ticket for that, just so we can get this in.

PR #1230, opened 2024-08-21
update search bar background

### LINKED ISSUE #1228: update search bar background
Make search bar background Accent/Fog/Light, height Hug (79px)

dnoneill:
closes #1228

### DIFF
--- a/app/assets/stylesheets/earthworks.css
+++ b/app/assets/stylesheets/earthworks.css
@@ -1,5 +1,6 @@
 :root {
   --stanford-50-black: #979694;
+  --stanford-fog-light-rgb: 244, 244, 244;
 }
 
 /* Selects both #main-content and navbar containers */
@@ -120,6 +121,11 @@
   border: 1px solid #ABABA9;
   padding: .5rem;
   border-radius: .25rem;
+  background-color: white;
+}
+
+.navbar-search {
+  --bs-light-rgb: var(--stanford-fog-light-rgb);
 }
 
 .search-query-form input:focus {


app/assets/stylesheets/earthworks.css
```diff
@@ -120,6 +121,12 @@
   border: 1px solid #ABABA9;
   padding: .5rem;
   border-radius: .25rem;
+  background-color: white;
```
jcoyne, line 124: Can you explain why this is needed?  It already appears white to me.

dnoneill: Without the background set. Because we aren't using the input border and putting our own on it needs a background color because without it inherits from the navbar-search css. 
Screenshot 2024-08-21 at 12 49 56 PM

jcoyne: I see a new version of geoblacklight.

Can you instead set `--bs-bg-opacity: 0`?

package.json
```diff
@@ -4,7 +4,7 @@
     "vite-plugin-ruby": "^5.0.0"
   },
   "dependencies": {
-    "@geoblacklight/frontend": "^5.0.0-alpha.3",
+    "@geoblacklight/frontend": "^5.0.0-alpha.9",
```
jcoyne, line 7: I think it's helpful to say that you are upgrading GBL in the PR title.

app/assets/stylesheets/earthworks.css
dnoneill: Doesn't work. Neither does --bs-navbar-color: white.

jcoyne: I've pulled this branch and I still see that background.

jcoyne: Ah, I see this is just inside the input-group.  Is there a `.search-query-form` that is not within the `.navbar-search`.  I can only see this being important in the latter case right now.

app/assets/stylesheets/earthworks.css
```diff
@@ -120,6 +121,12 @@
...
+}
+
+.navbar-search {
+  --bs-light-rgb: var(--stanford-fog-light-rgb);
+  margin-bottom: 2rem;
```
jcoyne, line 129: In other projects we've moved this spacing to `#main-container` as you often want spacing from the header even if it's not the `.navbar-search` that's right above it.

jcoyne: In fact https://github.com/sul-dlss/earthworks/blob/ff836cc0e8315c2fc81662809588b281ead86aba/app/assets/stylesheets/earthworks.css#L8 is the thing that is removing the default padding from the main container.

jcoyne: I think https://github.com/sul-dlss/earthworks/pull/1232 will obviate the need for this margin.

dnoneill: Looks like it did. Thanks.

hudajkhan (approved):

PR #1231, opened 2024-08-21
on show page, text align fields left

### LINKED ISSUE #1220: Left align metadata labels on record view
Metadata labels should be left-aligned so they can be scanned more quickly/easily. The record view should be adjusted per the following - https://www.figma.com/design/nkgg905s5FyXoyPZjVMco0/EarthWorks?node-id=911-24473&t=wbfBPbfvy5YnCPZw-4.

dnoneill:
Tried multiple ways to pass a class to <%= metadata %> in the document component. Couldn't get it to work.

closes #1220

### DIFF
--- a/app/assets/stylesheets/earthworks.css
+++ b/app/assets/stylesheets/earthworks.css
@@ -279,3 +279,10 @@ label.toggle-bookmark {
 .leaflet-container {
   background-color: #d4dadc!important;
 }
+
+/* Override https://github.com/projectblacklight/blacklight/blob/96a3a85c614ac1a4283c102f59fba05edf3c1cd0/app/assets/stylesheets/blacklight/_balanced_list.scss#L9-L11
+.dl-invert dt text-align: right;
+*/
+.dl-invert dt {
+  text-align: left;
+}


app/assets/stylesheets/earthworks.css
```diff
@@ -164,3 +164,7 @@ Uncomment lines 142-144
   --bs-pagination-focus-bg: none;
   --bs-pagination-focus-box-shadow: none;
 }
+
+.document-metadata dt, .document-metadata dd {
```
jcoyne, line 168: Are you trying to remove the `.dl-invert` styles here? If so, can you init your DocumentMetadataComponent without that class? https://github.com/projectblacklight/blacklight/blob/a8b5a3d3a2e8243c8a875d40528eae5c106d83bd/app/components/blacklight/document_metadata_component.rb#L12

dnoneill: not remove, just add text-start to those, but yes I would want DocumentMetadataComponent to have another class added.

jcoyne: I love additive styles rather than subtractive ones.

dnoneill: I think that works. I wasn't sure if really made sense to do a component override for such a small thing that could be fixed with css.

dnoneill: @jcoyne .dl-invert has more css than just text-align. What if we override the class instead?

jcoyne: The color? https://github.com/projectblacklight/blacklight/blob/4934e604e4538d22afff635579c8d11f9e23aba4/app/assets/stylesheets/blacklight/_balanced_list.scss#L4.   Ya'll can do what you want, but when maintaining an app, I find that CSS overrides are very hard to understand what they are trying to do compared to just applying classes that add style.

PR #1235, opened 2024-08-22
Add custom DocumentMetadataComponent to remove dl-invert class

jcoyne:


config/environments/development.rb
```diff
@@ -69,7 +69,7 @@
   # config.i18n.raise_on_missing_translations = true
 
   # Annotate rendered view with file names.
-  # config.action_view.annotate_rendered_view_with_filenames = true
+  config.action_view.annotate_rendered_view_with_filenames = true
```
dnoneill, line 72: What does this do?

app/controllers/catalog_controller.rb
```diff
@@ -48,6 +48,7 @@ class CatalogController < ApplicationController
 
     config.show.sidebar_component = Earthworks::Document::SidebarComponent
     config.show.document_component = Earthworks::DocumentComponent
+    config.show.metadata_component = DocumentMetadataComponent
```
dnoneill, line 51: This should probably be Earthworks::DocumentMetadataComponent

config/environments/development.rb
jcoyne: It shows you in HTML comments in the development server which views are being rendered.

app/controllers/catalog_controller.rb
jcoyne: I don't understand why we'd bother with namespaces in an application. That makes sense in a library, but for local stuff it just gets wordy.

marlo-longley: I'd be down to change that, that makes sense. We have ~12 components with the namespace right now though. Consistency?

jcoyne: https://github.com/sul-dlss/earthworks/pull/1249

PR #1238, opened 2024-08-22
Expand collapse 3

marlo-longley:
An approach w more stimulus

Builds on a would replace #1234

### DIFF
--- a/app/javascript/controllers/expandcollapse_controller.js
+++ b/app/javascript/controllers/expandcollapse_controller.js
@@ -1,16 +1,29 @@
 import { Controller } from "@hotwired/stimulus"
 
 export default class extends Controller {
-  static outlets = [ "description" ]
-  toggleDescriptions(evt) {
-    const activeClass = 'active'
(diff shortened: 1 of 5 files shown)


marlo-longley:
@jcoyne is this what you had in mind? Thanks.

app/javascript/controllers/expandcollapse_controller.js
```diff
@@ -0,0 +1,29 @@
+import { Controller } from "@hotwired/stimulus"
+
+export default class extends Controller {
+  static outlets = ["description"]
+  static targets = ["expandButton", "collapseButton"]
+
+  expandDescriptions() {
+    this.descriptionOutletElements.forEach(function (element) {
```
jcoyne, line 8: I think this is way better! However, I'd use `descriptionOutlets` and call `expand` on each controller.  Then the description controller can decide how to implement expand (in this case it will remove the `collapse` class.

marlo-longley: Nice, thanks. Done!

jcoyne (approved):

PR #1239, opened 2024-08-22
Update path to geoblacklight stylesheet

marlo-longley:


jmartin-sul (approved):
same change worked for me locally to get rid of a `Can't find stylesheet to import` error when running locally.

Issue #1110, opened 2024-07-31
Solr Synonyms: Terms with spaces


hudajkhan:
As mentioned in #1109 , synonym lines in the Solr config file may have multi-word terms where the words are separated by a space.  For multi-way expansion (i.e. lines with commas where a search for any term on the line means ORing the search with all the other terms on the line), these terms are not parsed correctly.  For example, the string "Public institutions, buildings" will lead to a search for "public" OR "institutions" OR "buildings" instead of "public institutions" OR "buildings"

hudajkhan:
The best way to handle multi-term synonyms appears to be to use the SynonymGraphFilterFactory: https://solr.apache.org/guide/solr/latest/indexing-guide/filters.html#synonym-graph-filter .  This filter is also currently in use with SearchWorks, vt arclight, etc. although there we can see this is applied at both index and query time, while our default configuration just used query time analysis.

hudajkhan:
Closed by https://github.com/sul-dlss/sul-solr-configs/pull/342

hudajkhan:
Will close once the changes have been deployed to production as a final step (this is a step that we must take after the GitHub merge so the changes are actually in the Solr collection)

hudajkhan:
Changes have been deployed to production.

PR #1240, opened 2024-08-22
Update synonyms: clean up synonyms file and remove 

### LINKED ISSUE #1109: Solr Synonyms: Review file to see if we want to make any changes to the terms we are searching for
Synonym file: https://github.com/sul-dlss/earthworks/blob/main/config/solr_configs/synonyms.txt

What these lines mean:

- For lines that have only commas, when a query (in the text field) is made for any term on the line, the query expands to include all the terms on the line.   For example, "Harbors,Pier,Quay,Warf" means searching for "Quay" will search for Quay and also for Pier and for Warf (i.e. like an OR search: Harbors OR Pier or Quay or Warf). 
- For lines that have "=>", the query will replace whatever is on the left with a search for whatever terms are on the right.  For example, "Invertebrates => Invertebrates,Lobsters,Mosquitoes,Shellfish" means that a search for Invertebrates will translate to Invertebrates OR Lobsters OR Mosquitoes OR Shellfish. 

The LCSH and ISO sections were retrieved from OGP's repo.  It would be useful to review these to ensure that we want the behavior that will result from the inclusion of these terms. 

In addition, lines that have terms with spaces (e.g. "Public institutions") will break those terms into their individual words at analysis time.  This is not behavior that we want so a separate issue will be created to handle this.

hudajkhan:
Closes #1109 .
Links to #1110 . 

What this pull request does:

- Synonyms text file: Removes the word "chemists" from the file as @kimdurante determined that this was word should not be included (see #1109 ).  Also removes extra spaces and many duplicate terms I saw (on the same line where they are not necessary). 
- Solr schema file: Change the synonym configuration for the "text_en" field type to use "SynonymGraphFilterFactory" instead.  SynonymFilterFactory has been deprecated, and SynonymGraphFilterFactory can handle multi-term synonyms. 
- For reference, see: https://solr.apache.org/guide/solr/latest/indexing-guide/filters.html#synonym-graph-filter which states "This filter is a replacement for the Synonym Filter, which produces incorrect graphs for multi-token synonyms.".
- Also for reference, see https://lucidworks.com/post/multi-word-synonyms-solr-adds-query-time-support/ . Note the three steps: Using SynonymGraphFilterFactory in the query analyzer, setting sow (i.e. split on whitespace) to false (this setting appears to be set to false by default in current/recent Solr versions), and setting expand to true.  Two of these were already in place for our Solr configuration. 

Testing:
To test this on my own local machine (which has its own local Solr), I set up three documents.  One had "dolphins" in the title and description, one had "mammals" in the title and description, and one had "marine mammals" in the title description.  The synonym line maps "dolphins" to "marine mammals".  

- When I searched for dolphins, I got both the document with the dolphins title as well as the marine mammals title, but NOT the "mammals" title.  This behavior appears correct, because the synonym "marine mammals" is a multi-term synonym and we want the entire term to match and not just "marine" or "mammals".  
- When I searched for "marine mammals", I got the dolphins document and the marine mammals document.
- When I searched for "mammals", I got both the "mammals" document and the "marine mammals" document.  

To test this out in your local app, you can spin up Solr using the app config (e.g. solr_wrapper or docker). You can look at the analysis dashboard section for the text field and see which terms are picked up as synonyms.  You can also pick a line from the synonyms file and pick two terms; set up two test documents with a term each in the title or description; and then test to see if the search works the way you intended.


marlo-longley:
@hudajkhan yay!
Questions:
- Why are these css changes in this PR?
- Are you able to add any of the logic you describe as testing steps into our specs? Say in `search_spec.rb`? Or maybe this isn't neccessary.

marlo-longley:
Also, to close out the ticket we would have to push these changes into `sul_solr_configs`, yes? Just checking.

hudajkhan:
Good catch @marlo-longley ! There shouldn't be any CSS changes, not sure what happened there.

hudajkhan:
@marlo-longley I will look into setting up a test.

hudajkhan:
Yes, to close out the synonym tickets, the relevant sul-solr-config PRs to stage and UAT should be set up. I'll refer to those PRs in the tickets.

hudajkhan:
@marlo-longley I added the test. Thanks for the suggestion!

marlo-longley (approved):

PR #1242, opened 2024-08-22
override dropdown button to add a new different class

### LINKED ISSUE #1207: update buttons in search to blue color
see comments here: https://github.com/sul-dlss/earthworks/pull/1190

peetucket:
Fixes #1207 - override dropdown button component from blacklight so we can set a different class

(note: requires main of blacklight until there is a new release since we depend on https://github.com/projectblacklight/blacklight/pull/3280/files and https://github.com/projectblacklight/blacklight/pull/3267 which are not in a current release)

Screenshot 2024-08-22 at 12 19 05 PM


app/components/blacklight/system/dropdown_button_component.rb
```diff
@@ -0,0 +1,18 @@
+# frozen_string_literal: true
+
+module Blacklight
+  module System
+    class DropdownButtonComponent < Blacklight::Component
+      def initialize(label:, classes: %w[btn btn-outline-primary dropdown-toggle])
+        @classes = classes
```
peetucket, line 7: change is here ... it uses `btn-outline-primary` instead of `btn-outline-secondary`

peetucket:
Note: we could create a new `Earthworks::System::DropdownComponent` and sublcass from the Blacklight component, so we would only update the class instead of duplicating the component here, but given the nature of how this component is rendered in blacklight, you'd then have to override and subclasses a couple other classes and a view too.

app/components/blacklight/system/dropdown_button_component.rb
jcoyne: Can we just extend Blacklight component here and change the default param, and call `super`?

jcoyne: I see that you've completely replaced the BL module.  I think overriding it in our root namespace might be more understandable.

peetucket: Can you give me an example of how you imagine this working? This was the approach I started with, but I found I had to override/sublcass two other BL components and a view in order to get to to render the `Earthworks` version of the components, and it felt a bit convoluted for a single class change.  Maybe I'm missing something.

jcoyne: I'm thinking https://github.com/projectblacklight/blacklight/pull/3289 might help.  Once that's in. I can look at https://github.com/projectblacklight/blacklight/issues/3290 which should make this trivial.

jcoyne: So, https://github.com/projectblacklight/blacklight/pull/3296 now will allow this component to be configured.

peetucket: Thanks Justin.  For this PR, I just need to override the class for the button that is rendered, which is in`DropdownButtonComponent`: https://github.com/projectblacklight/blacklight/blob/main/app/components/blacklight/system/dropdown_button_component.rb  

It looks like the reference to the button component is still hardcoded within the `DropdownComponent`: https://github.com/projectblacklight/blacklight/blob/main/app/components/blacklight/system/dropdown_component.rb#L6

If I create new a Earthworks `DropDownComponent`, subclass it from the Blacklight component, use the new blacklight config to tell it to use this new Earthworks::DropdownComponent, then create a new Earthworks specific `DropdownButtonCompontent` (also subclassed from the blacklight one) so that I can tell it render with my new css class (e.g. just do this in my subclass `DropDownComponent`: `renders_one :button, Earthworks::System::DropdownButtonComponent`), it complains of a duplicate slot:

```
Earthworks::System::DropdownComponent declares the button slot multiple times.
To fix this issue, choose a different slot name.
```

peetucket: Maybe if this component https://github.com/projectblacklight/blacklight/blob/main/app/components/blacklight/system/dropdown_component.rb#L6 was configurable in blacklight in a similar way to what you did in https://github.com/projectblacklight/blacklight/pull/3296, I could create a new Earthworks `DropdownButtonComponent`, subclass from the blacklight one and update the css class as needed.

jcoyne: I'm not sure I understood, but presently, BL allows you to configure which DropdownComponent you want to use.

marlo-longley:
@peetucket looks good, but there a still red buttons around. Should we open new tickets specifically for those so that we can move on? Or would you like to address that here?

Screenshot 2024-08-30 at 16 17 25

peetucket:
Yeah, this PR only addresses the specific buttons mentioned in the ticket.  Unless it is not acceptable to release/deploy this change without the other buttons changing, I'd suggest we find the other buttons that are still red and ticket those separately (assuming that is the intended change, which I agree it makes sense for it to be).  The way the buttons are changed may end up being different than what was ultimately done in this PR too.

marlo-longley (changes requested):
We stopped namespacing EW components in https://github.com/sul-dlss/earthworks/pull/1249 -- could you update?

marlo-longley:
I opened tickets for the other buttons as requested.

peetucket:
Namespace of component removed

peetucket:
Closed in favor of #1348 instead of doing a messy rebase here
Same commits

PR #1249, opened 2024-08-22
Stop namespacing components

jcoyne:


thatbudakguy:
@jcoyne needs rebase. if we can manage to get this without conflicts I'm good to merge it, but might be tricky...

PR #1267, opened 2024-08-26
Reduce the app header vertical padding slightly

thatbudakguy:
This matches Figma (or at least gets closer).

Before:
Screenshot 2024-08-26 at 14 24 33

After:
Screenshot 2024-08-26 at 14 24 50

Figma:
Screenshot 2024-08-26 at 14 24 17

### DIFF
--- a/app/components/blacklight/top_navbar_component.html.erb
+++ b/app/components/blacklight/top_navbar_component.html.erb
@@ -161,7 +161,7 @@
       <div
         class="col-md-8 d-flex justify-content-center justify-content-md-start"
       >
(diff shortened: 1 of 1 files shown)


jcoyne:
@thatbudakguy if this is what @dbranchini wants. Can you do this also in the component-library, so we don't have to keep correcting it going forward?

dbranchini:
Yes, and looks good. Thanks!

marlo-longley (approved):

thatbudakguy:
Also opened https://github.com/sul-dlss/component-library/pull/34.

PR #1272, opened 2024-08-28
911 zoom level

### LINKED ISSUE #911: Set zoom level on landing page maps so that only one world map is shown

dnoneill:
closes #911 
Screenshot 2024-08-28 at 2 59 00 PM

### DIFF
--- a/app/assets/stylesheets/earthworks.css
+++ b/app/assets/stylesheets/earthworks.css
@@ -264,3 +264,8 @@ label.toggle-bookmark {
   margin-right: .25rem;
   vertical-align: middle;
 }
+
+/* Have to use !important because we load leaflet.css after earthworks.  */
+.leaflet-container {
+  background-color: #d4dadc!important;
+}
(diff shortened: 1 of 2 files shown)


jcoyne:
Will this be difficult for users trying to look at Beringia (area in Russia that is bifurcated by the edge)? What happens if you zoom in there?

dnoneill:
@jcoyne Seems to work okay.

hudajkhan:
Works as intended.  
On UAT without this merge:
Screenshot 2024-08-28 at 5 33 58 PM

After
Screenshot 2024-08-28 at 5 36 23 PM

hudajkhan (approved):

PR #1273, opened 2024-08-28
Double search

dnoneill:
Screenshot 2024-08-28 at 5 08 13 PM


jcoyne:
This seems like an unusual and unexpected way of customizing blacklight. Have you considered setting the search_bar_component class instead? (overriding Blacklight::SearchBarComponent)?

dnoneill:
Done.

app/controllers/catalog_controller.rb
```diff
@@ -40,6 +40,7 @@ class CatalogController < ApplicationController
 
     config.index.document_component = Earthworks::SearchResultComponent
     config.index.title_field = Settings.FIELDS.TITLE
+    config.index.search_bar_component = Earthworks::SearchBarComponent
```
jcoyne, line 43: Do we want to use this component for the show view too?

dnoneill: I don't see a config.show.search_header_component in blacklight and it looks like the show page is using the custom search_header_component

jcoyne (approved):

PR #1291, opened 2024-08-29
Use additive style rather than subtractive

jcoyne:
This gives slightly more content area on small screens (bootstrap default) and avoids the need for an extra rule


thatbudakguy:
needs rebase

thatbudakguy (approved):

Issue #1277, opened 2024-08-28
Remove extra border around facets on mobile view


dbranchini:
Each facet has a border, so we do not need to wrap all of them with another border. It's only happening on mobile view. Also, I think we need to center "Limit your search" and the "Show/Hide facets" button vertically. And lastly, if possible, the button should be a primary outline button instead of the default gray. 

<img src="https://github.com/user-attachments/assets/8e14f67f-136f-492a-bbb9-684e9ab95897" width="250">

<img src="https://github.com/user-attachments/assets/c1479cdf-fe2c-4e6b-a7ed-834147aa91c0" width="250">

hudajkhan:
The border line appears to come in from Blacklight: https://github.com/projectblacklight/blacklight/blob/main/app/assets/stylesheets/blacklight/_facets.scss#L52

hudajkhan:
For the border color for the button, would you mind providing a similar example or the color you like? I assume you want btn-primary-outline (or similar) but just want to confirm.  

Also, I am applying the spacing changes only to the mobile version.  Is the correct? Or would you like to change the border color for all window sizes, not just mobile?

hudajkhan:
The PR above addresses the section border and spacing, but not the button itself.  It might be good to do a different PR overriding the button as a whole/using the component as opposed to just CSS changes.

hudajkhan:
Also, the button doesn't exist in the desktop version, so that answers one of my questions above.

dbranchini:
It should use the primary blue outline button seen here - https://sul-dlss.github.io/component-library/button/. Pick a size that's closest to the button that's there now.

hudajkhan:
Thanks.  For reference, the code for the "show/hide facets" button is here: https://github.com/projectblacklight/blacklight/blob/main/app/components/blacklight/response/facet_group_component.html.erb

jcoyne:
This may have been an inadvertent change in Blacklight https://github.com/projectblacklight/blacklight/commit/32373178d88f529855adbb73df3630865a8e53ba. Can we fix it there?

jcoyne:
Actually, it's just easer to set `  --bl-facets-smallish-border: none;`

hudajkhan:
Closed by #1309

PR #1309, opened 2024-08-29
Removing border from facet section, changing spacing of heading and button

hudajkhan:
Addresses #1277 

To remove the border around the facet section for the mobile view, we override the border value for .facets-toggleable-md.  To vertically align the heading and the button, we remove some padding and margin values that were pushing the text further below.


app/assets/stylesheets/earthworks.css
```diff
@@ -285,3 +285,14 @@ label.toggle-bookmark {
   --bl-facet-active-item-color: rgb(var(--stanford-palo-alto-rgb));
   border-color: rgb(var(--stanford-palo-alto-rgb));
 }
+
+/* Override mobile view for facets section in Blacklight */
+@media (max-width: 991.98px) {
+  .facets-toggleable-md {
+    border: none;
+    padding-top: 0px;
```
dnoneill, line 293: I would change this to 
.facets {
  --bl-facets-smallish-border: none;
  --bl-facets-smallish-padding: 0;
}

And I think you can get rid of the @media

hudajkhan: I changed the above, but couldn't figure out how to get facets-heading to change in a similar fashion (I tried --bl-facets-smallish-margin-bottom but the .facets-heading margin-bottom value of 0.5 rem kept overriding that)

app/assets/stylesheets/earthworks.css
```diff
@@ -317,3 +317,29 @@ label.toggle-bookmark {
...
+    margin-bottom: 0;
+  }
+}
+
+/* Hide/show button on mobile */
+.facets-header .navbar-toggler {
+  --bl-facets-smallish-border: none;
+  --bs-navbar-toggler-border-color: var(--stanford-digital-blue);
+  --bs-navbar-color: var(--stanford-digital-blue);
+  color: var(--stanford-digital-blue);
```
jcoyne, line 339: is this `color` necessary? This rule should apply:
```
.navbar-toggler {
...
	color: var(--bs-navbar-color);
}
```

app/assets/stylesheets/earthworks.css
```diff
@@ -317,3 +317,29 @@ label.toggle-bookmark {
...
+}
+
+.facets-header .navbar-toggler:hover {
+  color: var(--stanford-digital-blue-dark);
```
jcoyne, line 343: We should make this settable in blacklight: https://github.com/projectblacklight/blacklight/blob/69373f202753f04ec6aca179fdf8fa01248a9edf/app/assets/stylesheets/blacklight/_facets.scss#L22-L25

app/assets/stylesheets/earthworks.css
hudajkhan: For whatever reason, it doesn't work with just the "--bs-navbar-color" definition.The "color" property is set explicitly elsewhere, so --bs-navbar-color isn't the only thing determining the color.

app/assets/stylesheets/earthworks.css
hudajkhan: Do you mean, for example, make a variable specifically for the hover color in Blacklight, so we can just override that  variable value?

jcoyne: Exactly that.

hudajkhan:
To make the rebase to main easier, and to get this moving along, I am going to close this PR and open up two new ones: 
one for the outline around the facet section and vertical alignment and one for changing the button to a primary outline button.

PR #1315, opened 2024-08-29
Hierarchical date facet

### LINKED ISSUE #1136: Add hierarchy to the year facet
The blacklight_range_limit slider has accessibility issues. We're looking at implementing a hierarchical facet, similar to vt-arclight:

Screenshot 2024-08-08 at 09 45 59

To get started, see [the implementation there](https://github.com/sul-dlss/vt-arclight/blob/8206bfb9f2fb1c4b52a8682d6efbd97d0a579b0d/app/controllers/catalog_controller.rb#L136-L140).

jmartin-sul:
closes #1136

needs https://github.com/sul-dlss/searchworks_traject_indexer/pull/1521 to populate the field for the facet.

Screenshot 2024-09-05 at 11 45 09 PM

the alignment seems a little weird on the facet removal button (`x`) in that screenshot from my local instance.  but this seems like an improvement over the current facet, so i wonder if we could merge this and refine the CSS if that's the only issue, and the fix isn't totally obvious?

if removing the blacklight_range_limit gem isn't desired afterall, i left that as a separate commit, so easy enough to revert that part or split it out into another PR.


spec/features/blacklight_customizations/hierarchical_date_facet_spec.rb
```diff
@@ -0,0 +1,24 @@
+require 'rails_helper'
+
+describe 'Hierarchical date facet' do
+  before do
+    visit root_path
+    click_link_or_button 'search'
+  end
+
+  it 'includes an expandable date facet' do
+    expect(page).to have_content('1 - 18 of 18')
```
marlo-longley, line 10: In a recent PR of mine (https://github.com/sul-dlss/earthworks/pull/1337) we decided that hardcoding fixture counts into the tests was going to cause annoyance in the future.
This is more targeted because it's counting a field that will probably not be updated frequently. But if there was a way to test this without using counts, I'd go for that. If not, I think it's fine.

app/components/blacklight/date_facet_hierarchy_component.rb
```diff
@@ -0,0 +1,47 @@
...
+      end
+    end
+
+    private
+
+    # Build out the tree hierarchy (century -> decade -> year) for
+    # a facet and return the top-level facet items
+    #
+    # @return [Array<Blacklight::Solr::Response::Facets::FacetItem>]
+    def facet_item_tree_hierarchy_from_year_list
```
marlo-longley, line 18: I have read through this method and understand what you're doing but it is the trickiest part of the code. It might benefit from a comment about what format input is expected, and also about the different cases you handle along the way.

app/components/blacklight/date_facet_hierarchy_component.rb
```diff
@@ -0,0 +1,47 @@
...
+      tree = {}
+
+      @facet_field.paginator.items.each do |item|
+        century_str, decade_str, year_str = item.value.split(':')
+
+        tree[century_str] ||= Blacklight::Solr::Response::Facets::FacetItem.new(value: century_str, items: [], hits: 0)
+
+        next unless decade_str
+
+        decade_facet = tree[century_str].items.find do |decade_item|
```
marlo-longley, line 28: I had a little trouble following when this case would ever be true, because 1) you explicitly set ` tree[century_str].items` to an empty array on line 24, and 2) you've already split the string on `:` on line 22. 

I concluded that this `decade_facet` could only exist if you'd already created it during another loop.

spec/features/blacklight_customizations/hierarchical_date_facet_spec.rb
jmartin-sul: ah yeah, good point, will think for a bit about trying to get out of using counts...

app/components/blacklight/date_facet_hierarchy_component.rb
jmartin-sul: good call, will do

app/components/blacklight/date_facet_hierarchy_component.rb
jmartin-sul: yeah, that's right, this code is looking to update an existing instance if it finds one, or to add that decade for the first time if need be.  will add a comment...

thatbudakguy (approved):
Tested on stage and ready to merge, pending response to feedback

spec/features/blacklight_customizations/hierarchical_date_facet_spec.rb
jmartin-sul: ok, so i wanted to keep somewhat precise search behavior expectations, to confirm that faceting wasn't happening when it shouldn't but also that faceting narrowed results appropriately when applied.  but the worry about test fragility and annoyance totally makes sense.

so given that the main worry in the other PR's discussion seemed to be about frequent annoyance when adding unrelated fixtures, and since it seemed less common that we'd regularly add or remove fixtures in the 1990-1999 date range, i tried to find a middle ground: i kept the hardcoded count for the post-facet results; and for the prefacet expectations, i just looked for the initial total result count, and made sure that stayed unchanged when expanding (but not choosing) a facet in the tree.

hopefully that's ok, but happy to revisit if that still feels too brittle!

app/components/blacklight/date_facet_hierarchy_component.rb
jmartin-sul: i may have gone too far in the other direction, but i added what should hopefully be a thorough explanation of expected input and output structures 😅

app/components/blacklight/date_facet_hierarchy_component.rb
jmartin-sul: added some inline comments that try to explain what the conditionals mean in the larger context (of both the current loop iteration and the prior loop iterations), as well as a couple example strings that would meet the condition for skipping to the next loop iteration.

spec/features/blacklight_customizations/hierarchical_date_facet_spec.rb
```diff
@@ -0,0 +1,24 @@
...
+  end
+
+  it 'includes an expandable date facet' do
+    result_count_display = find('span.page-entries').text
+    expect(result_count_display).to match(/1 - \d+ of \d+/)
+
+    click_link_or_button 'Date'
+
+    find_link('1900-1999').ancestor('li.treeitem').first(:button).click
+    expect(page).to have_content(result_count_display) # have only expanded the facet for 1900s, haven't filtered yet
```
jmartin-sul, line 16: try to keep the test from breaking when unrelated fixtures are added, while confirming that simply expanding a tree item doesn't facet on that node immediately.

spec/features/blacklight_customizations/hierarchical_date_facet_spec.rb
```diff
@@ -0,0 +1,24 @@
...
+    expect(page).to have_css('li.treeitem ul li.treeitem', text: /1990-1999/, visible: :visible)
+    click_link_or_button '1990-1999'
+    expect(page).to have_content('1 - 3 of 3')
```
jmartin-sul, line 19: hopefully we won't regularly be adding or removing fixtures that fall in the 1990-1999 time range, but happy to revisit if thus hardcoded count still feels too brittle.

app/components/blacklight/date_facet_hierarchy_component.rb
```diff
@@ -0,0 +1,72 @@
...
+    # an object that covers 1996 will have a field that looks like:
+    # ["1900-1999", "1900-1999:1990-1999", "1900-1999:1990-1999:1996"]
+    # See the date_hierarchy_sm field in this repo's fixture data or in
+    # searchworks_traject_indexer for more examples/explanation.
+    #
+    # Output: The tree that's built is one of nested facet items.  The top level is a list of
+    # FacetItem instances, one per century covered.  Each century's #items list has one FacetItem for
+    # each decade covered by that century.  And each decade FacetItem#items instance has one entry for each
+    # year covered by that decade.  Thus, a tree entry for each element in @facet_field.paginator.items,
+    # allowing faceting for each level of the hierarchy.
```
jmartin-sul, line 31: happy to edit down if this is overkill 😅

jmartin-sul (commented):
@marlo-longley i think this is ready for another look, thanks for the feedback!

i pushed the changes as two new commits, both for easier review, and for easier undo if more iteration is needed.  happy to squash before merge if approved and minimizing the number of new commits is desired.  (but the commits also seem like logically separate chunks of development, so i'm also fine leaving as is)

app/components/blacklight/date_facet_hierarchy_component.rb
marlo-longley: Ty!!

spec/features/blacklight_customizations/hierarchical_date_facet_spec.rb
marlo-longley: This seems fine.

marlo-longley (approved):
This looks good to me.

Issue #1279, opened 2024-08-28
Search input on search results and record detail view


dbranchini:
The search input looks larger in both width and height than designed. It's about 520 pixels on XXL, XL, L, M, S viewport sizes and full width on XS [as designed](https://www.figma.com/design/nkgg905s5FyXoyPZjVMco0/EarthWorks?node-id=335-13898&t=Mk1sK8D8gfmkr0Bx-4). It's also 47 pixels in height. Here's the specs in Figma:

Search icon - 24 pixels (width and height)
Text - 16 pixels 
Search button - 31 pixels (height) 

Search button is the largest in height, then there's 8 pixels padding top/bottom so that ends up being 47 pixels. I know it's done differently in code, so I don't expect it to match perfectly, just wanted to provide the full specs.

marlo-longley:
#1317 fixed most of this except for the magnifying class icon sizing.

PR #1317, opened 2024-08-30
Change padding and alignment for search bar

marlo-longley:
Part of #1279

Does not cover the resizing of the magnifying glass icon; I will wait until we merge code around the `search_component` for that. Not closing the ticket yet.

Screenshot 2024-08-30 at 10 00 16
Screenshot 2024-08-30 at 10 00 22

### DIFF
--- a/app/assets/stylesheets/earthworks.css
+++ b/app/assets/stylesheets/earthworks.css
@@ -128,7 +128,7 @@
 }
 
 .search-q {
-  padding-left: 2rem;
+  padding: 0 0 0 2rem;
   background: url("data:image/svg+xml,%3Csvg xmlns='http://www.w3.org/2000/svg' width='20' height='20' fill='currentColor' class='bi bi-search' viewBox='0 0 16 16'%3E%3Cpath d='M11.742 10.344a6.5 6.5 0 1 0-1.397 1.398h-.001q.044.06.098.115l3.85 3.85a1 1 0 0 0 1.415-1.414l-3.85-3.85a1 1 0 0 0-.115-.1zM12 6.5a5.5 5.5 0 1 1-11 0 5.5 5.5 0 0 1 11 0'/%3E%3C/svg%3E") no-repeat left;
   background-size: 1.5rem;
   outline: none;
@@ -138,22 +138,33 @@
   border: none;
 }
 
-.search-query-form, .navbar-search .search-query-form{
+.search-query-form {
   border: 1px solid #ABABA9;
   padding: .5rem;
   border-radius: .25rem;
   background-color: white;
+  width: 100%;
+  @media (min-width: 576px) {
+    max-width: 520px;
+  }
 }
 
 .navbar-search {
   --bs-light-rgb: var(--stanford-fog-light-rgb);
 }
 
+.navbar-search > .container-fluid {
+  @media (max-width: 767.98px) {
+    justify-content: center;
+  }
+}
 .search-query-form input:focus {
   box-shadow: none;
 }
 .search-query-form .search-btn {
   border-radius: .25rem!important;
+  padding: 2px 6px;
+  align-items: center;
 }
 
 /* TODO: when we get a new blacklight release (v8.3.1)


app/assets/stylesheets/earthworks.css
```diff
@@ -134,26 +134,48 @@
...
+    justify-content: space-between;
+  }
+}
 .search-query-form input:focus {
   box-shadow: none;
 }
 .search-query-form .search-btn {
   border-radius: .25rem!important;
+  max-height: 31px;
+  align-items: center;
```
jcoyne, line 178: Why do we need this? I don't see it having any effect.

app/assets/stylesheets/earthworks.css
```diff
@@ -134,26 +134,48 @@
...
+    max-width: 520px;
+  }
 }
 
 .navbar-search {
   --bs-light-rgb: var(--stanford-fog-light-rgb);
 }
 
+.navbar-search > .container-fluid {
+  justify-content: center;
```
jcoyne, line 167: can this be rewritten as:
```
@media(max-width: 767px) {
  justify-content: center;
 }
 ```
 Then I think you can leave out the `justify-content: space-between`, right?

File: app/assets/stylesheets/earthworks.css [hunk 3]
```diff
@@ -134,26 +134,48 @@
...
-.search-query-form, .navbar-search .search-query-form{
+.input-group > .search-autocomplete-wrapper input {
+  padding-top: 0;
+  padding-bottom: 0
+}
+
+.search-query-form, .navbar-search .search-query-form {
   border: 1px solid #ABABA9;
-  padding: .5rem;
+  padding: 8px;
```
jcoyne, line 152: These are equivalent.  Can you revert this so the design will be relative to the base font height?

app/assets/stylesheets/earthworks.css
```diff
@@ -134,26 +134,48 @@
...
   border-radius: .25rem;
   background-color: white;
+  max-height: 47px;
```
jcoyne, line 155: It shouldn't be necessary to set this.  The height of this is governed by the height of the button and the height of `.search-btn` and `.search-q` If that is constrainted to 29, this is what you will get.

app/assets/stylesheets/earthworks.css
```diff
@@ -134,26 +134,48 @@
   outline: none;
 }
 
+.search-q .input-group {
+  padding: .5rem
```
jcoyne, line 138: This shouldn't be necessary since the `.search-query-form` (which contains this and only this element) has the same padding already.

app/assets/stylesheets/earthworks.css
```diff
@@ -134,26 +134,48 @@
...
+}
+
 .input-group > .search-autocomplete-wrapper {
   border: none;
 }
 
-.search-query-form, .navbar-search .search-query-form{
+.input-group > .search-autocomplete-wrapper input {
+  padding-top: 0;
+  padding-bottom: 0
```
jcoyne, line 147:
```suggestion
  padding-block: 0;
```

app/assets/stylesheets/earthworks.css
marlo-longley: With align-items:
Screenshot 2024-08-30 at 10 37 58


Without (text at base of btn):
Screenshot 2024-08-30 at 10 38 01

app/assets/stylesheets/earthworks.css
marlo-longley: I always default to min-width over max-width for a mobile first approach. Max-width is last resort if you can't do it another way.
I can see in this case with the GBL overrides why this could be a good approach.

app/assets/stylesheets/earthworks.css
marlo-longley: I don't want all 4 sides to be zero. I don't want to override the left and right padding.

app/assets/stylesheets/earthworks.css
marlo-longley: The ticket said the button should be 31. But I can let 2 px go!

marlo-longley:
Thanks @jcoyne -- cleaner now, revised.

app/assets/stylesheets/earthworks.css
jcoyne: So this is due to the fight between padding and max-height.  I think we can eliminate this if we also remove max-height and just set padding and font size appropriately.

jcoyne (approved):

app/assets/stylesheets/earthworks.css
marlo-longley: Whoops. I can try this still...

PR #1319, opened 2024-08-30
Revise copy for homepage

### LINKED ISSUE #1316: Copy for the home page
Geospatial Datasets
Data that includes location information (coordinates) and is natively usable in GIS software and services such as ArcGIS, QGIS, WFS, and WMS. Geospatial data types include vector data, such as Shapefile, GeoJSON, and GeoPackage, and raster data, such as GeoTIFF, grid files, and georeferenced maps.

Maps
Images of paper maps that have been scanned and/or georeferenced. These digital cartographic resources cover all parts of the world and include topographical maps, historical maps, nautical charts, thematic maps, and atlases.

marlo-longley:
Closes #1316

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -6,10 +6,10 @@ en:
     home:
       datasets:
         header: Geospatial data
-        description: Need copy for this ... contains over %{sets} datasets...
+        description: Data that includes location information (coordinates) and is natively usable in GIS software and services such as ArcGIS, QGIS, WFS, and WMS. Geospatial data types include vector data, such as Shapefile, GeoJSON, and GeoPackage, and raster data, such as GeoTIFF, grid files, and georeferenced maps.
       maps:
         header:  Scanned maps
-        description: Need copy for this ... contains over %{sets} maps...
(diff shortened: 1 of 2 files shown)


config/locales/en.yml
```diff
@@ -6,12 +6,12 @@ en:
     home:
       datasets:
         header: Geospatial data
-        description: Need copy for this ... contains over %{sets} datasets...
+        description: Data that includes location information (coordinates) and is natively usable in GIS software and services such as ArcGIS, QGIS, WFS, and WMS. Geospatial data types include vector data, such as Shapefile, GeoJSON, and GeoPackage, and raster data, such as GeoTIFF, grid files, and georeferenced maps.
       maps:
         header:  Scanned maps
-        description: Need copy for this ... contains over %{sets} maps...
+        description: Images of paper maps that have been scanned and/or georeferenced. These digital cartographic resources cover all parts of the world and include topographical maps, historical maps, nautical charts, thematic maps, and atlases.
```
dnoneill, line 12: If we are no longer using sets, we should update the code to not calculate the number of records.

marlo-longley: Hm. Let's confirm we won't want that anywhere? I will add as a standup topic. Good idea!

PR #1320, opened 2024-08-30
Force the footer to the bottom on a tall screen

### LINKED ISSUE #1290: Homepage main container does not expand to fill page
There's extra whitespace on the bottom. Think we might need some `flex` on this guy.

Image

jcoyne:
Fixes #1290

### DIFF
--- a/app/assets/stylesheets/earthworks.css
+++ b/app/assets/stylesheets/earthworks.css
@@ -4,6 +4,11 @@
   --stanford-palo-alto-rgb: 23, 94, 84;
 }
 
+#main-container {
+  /* force the footer to the bottom on a tall viewport */
+  min-height: calc(100vh - 323px);
+}
+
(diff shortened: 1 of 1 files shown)


app/assets/stylesheets/earthworks.css
```diff
@@ -4,6 +4,11 @@
   --stanford-palo-alto-rgb: 23, 94, 84;
 }
 
+#main-container {
+  /* force the footer to the bottom on a tall viewport */
+  min-height: calc(100vh - 323px);
```
marlo-longley, line 9: What is the source of 323?

jcoyne: Header height + footer height.

thatbudakguy:
@jcoyne the way I usually do this is #1352 because it avoids needing to calculate a pixel value. sorry for not specifying that in the ticket

PR #1324, opened 2024-08-30
update facets to col-lg-2

### LINKED ISSUE #1276: Change the Bootstrap columns on search results page
I'm mainly speaking about how this renders on a 16-inch Mac laptop (or larger), so I'm happy to hop on a call to cover different breakpoints together. We are currently rendering search results as follows:

- Facets sidebar -- 3 columns 
- Search results + map panel -- 9 columns
-- Search results -- 6 columns 
-- Map panel -- 6 columns

Proposed solution:
- Facets sidebar -- 2 columns 
- Search results + map panel -- 10 columns
-- Search results -- 7 columns 
-- Map panel -- 5 columns
 
This is what it looks like with the proposed solution:

Image

dnoneill:
closes #1276

### DIFF
--- a/app/helpers/earthworks_blacklight_helper.rb
+++ b/app/helpers/earthworks_blacklight_helper.rb
@@ -1,10 +1,22 @@
 module EarthworksBlacklightHelper
   include Blacklight::LayoutHelperBehavior
+  # Blacklight override for show main content
   def show_content_classes
     'col-lg-10 show-document row'
   end
 
+  # Blacklight override for show page sidebar
   def show_sidebar_classes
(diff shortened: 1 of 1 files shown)


app/helpers/earthworks_blacklight_helper.rb
```diff
@@ -8,6 +8,14 @@ def show_sidebar_classes
     'page-sidebar col-lg-2'
   end
 
+  def main_content_classes
```
jcoyne, line 11: Would you mind dropping comments on each method that states this is a Blacklight override.

jcoyne (approved):

Issue #1301, opened 2024-08-29
Bookmarks page


dbranchini:
Bookmarks was designed as follows - https://www.figma.com/design/nkgg905s5FyXoyPZjVMco0/EarthWorks?node-id=385-13343&t=Mxo8GLwgXQqO5i98-4.

![Image](https://github.com/user-attachments/assets/d276c039-275c-4345-9793-79fc8f367bf6)

It more closely matches the layout of the search results page. If it's not possible at this late stage, then we should move the "Clear bookmarks" next to the "Bookmarks" heading, and it should be a secondary red outline button. 

![Image](https://github.com/user-attachments/assets/f43aa4fb-e471-4238-bf94-c4b0edd0db27)

marlo-longley:
Right now the page is totally broken. The links to certain records do not work, and the "bookmarked" status next to them is incorrect.

@dbranchini -- I see you added "cite" and "email" in your Figma. Does this make sense for a search-results style page with multiple bookmarks? Which would be cited in this case?

dbranchini:
Oh, it was intended to be a way to see citations for all and/or email all records. It came up during user testing, but we also heard a lot more from users in terms of functionality, so if we'd rather address a more holistic redesign of bookmarks at a later time (for all apps), that makes sense to me.

PR #1332, opened 2024-08-30
Customize bookmarks page layout

marlo-longley:
Addresses (but may not close) https://github.com/sul-dlss/earthworks/issues/1301

### DIFF
--- a/app/assets/stylesheets/earthworks.css
+++ b/app/assets/stylesheets/earthworks.css
@@ -103,7 +103,7 @@
   } 
 }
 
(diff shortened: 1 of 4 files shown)


marlo-longley:
@dbranchini I made a start here, but didn't get to change all of the button alignment. How does this seem for now? Thanks.



Screenshot 2024-08-30 at 13 57 55

dbranchini:
Yes, this is good. Thanks! Later we can move some of the buttons.

PR #1346, opened 2024-08-30
1346 Make facet buttons blue

### LINKED ISSUE #1340: Map facet buttons blue not red
See "maps"

peetucket:
Fixes #1340 

Looks like this in PR:

Screenshot 2024-08-30 at 3 37 31 PM
Screenshot 2024-08-30 at 3 37 35 PM

### DIFF
--- a/app/assets/stylesheets/earthworks.css
+++ b/app/assets/stylesheets/earthworks.css
@@ -100,7 +100,7 @@
   /* center these widgets on small screens */
   @media (max-width: 575.98px) {
     justify-content: center;
-  } 
+  }
 }
 
 .search-widgets .btn, .clear-bookmarks {
@@ -120,7 +120,7 @@
     & > .page-links {
       padding-right: 0;
     }
-  } 
+  }
 }
 
 .sort-pagination, .pagination-search-widgets {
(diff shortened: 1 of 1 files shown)


app/assets/stylesheets/earthworks.css
```diff
@@ -313,3 +313,21 @@ label.toggle-bookmark {
   --bs-link-color-rgb: var(--stanford-digital-blue-dark-rgb);
   text-decoration: underline;
 }
+
+.constraint-value.btn-outline-secondary {
```
jcoyne, line 317: Secondary is defined as red in the component library.  Can you change the class to be btn-outline-secondary instead?

jcoyne (changes requested):

app/assets/stylesheets/earthworks.css
jcoyne: This is in https://github.com/projectblacklight/blacklight/blob/a4ec2d85eeb8d61bd2d462ccdfb5de1c50dba27d/app/components/blacklight/constraint_layout_component.html.erb#L2

jcoyne: That is provided by https://github.com/projectblacklight/blacklight/blob/a4ec2d85eeb8d61bd2d462ccdfb5de1c50dba27d/app/components/blacklight/constraint_component.rb#L7
and https://github.com/projectblacklight/blacklight/blob/a4ec2d85eeb8d61bd2d462ccdfb5de1c50dba27d/app/components/blacklight/constraints_component.rb#L26C20-L26C46
Which you can configure in https://github.com/projectblacklight/blacklight/blob/a4ec2d85eeb8d61bd2d462ccdfb5de1c50dba27d/lib/blacklight/configuration.rb#L187

jcoyne:
This seems to be in conflict with what is asked for in https://github.com/sul-dlss/earthworks/issues/1355

peetucket:
> This seems to be in conflict with what is asked for in #1355

Indeed it does -- I didn't see the figma design when doing this ticket.  Closing this PR and the associated ticket.

PR #1347, opened 2024-08-30
1343 Make expand/collapse buttons blue

### LINKED ISSUE #1341: Make expand/collapse search result buttons blue not red

peetucket:
Fixes #1341 - make expand/collapse buttons blue

With this PR:

Screenshot 2024-08-30 at 2 53 12 PM

### DIFF
--- a/app/assets/stylesheets/earthworks.css
+++ b/app/assets/stylesheets/earthworks.css
@@ -208,11 +208,10 @@ Uncomment lines 142-144
 
 .expand-collapse .btn-outline-primary {
   display: flex;
-  --bs-btn-hover-color: var(--stanford-digital-blue-dark);
-  --bs-btn-hover-border-color: var(--stanford-digital-blue-dark);
-  --bs-btn-active-color: white;
-  --bs-btn-active-bg: var(--stanford-digital-blue-darker);
-  --bs-btn-active-border-color: var(--stanford-digital-blue-darker);
+  &.active {
+    color: white;
+    background-color:var(--stanford-digital-blue-darker);
+  }
 }
 
 .expand-collapse i {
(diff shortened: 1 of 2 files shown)


app/assets/stylesheets/earthworks.css
```diff
@@ -208,10 +208,13 @@ Uncomment lines 142-144
 
 .expand-collapse .btn-outline-secondary {
```
jcoyne, line 209: It's really odd that we're changing secondary buttons to be blue.  The component library defines "primary" as the blue color.  Is there a reason we can't change the class?

jcoyne (changes requested):

app/assets/stylesheets/earthworks.css
```diff
@@ -206,12 +206,13 @@ Uncomment lines 142-144
...
 /* Styles for search result description expand/collapse */
 
-.expand-collapse .btn-outline-secondary {
+.expand-collapse .btn-outline-primary {
   display: flex;
-  &.active {
-    color: white;
-    background-color:var(--stanford-cardinal);
-  }
+  --bs-btn-hover-color: var(--stanford-digital-blue-dark);
```
jcoyne, line 211: None of these are necessary anymore, right? Isn't this the default?

jcoyne: See https://github.com/sul-dlss/component-library/blob/355885ec11bcaca6d601bb7eae3b44251e44fa0d/styles/button.css#L37

peetucket: I just reviewed, and you are right that several are default in the component library and can be removed.  But the active button turning solid blue isn't the default and I think it's what is requested.  So I think a couple overrides are still relevant.

peetucket: Actually, was able to adjust the existing `&.active` definition to be blue instead of red.  Seemed to do it.

jcoyne (approved):

PR #1352, opened 2024-09-03
Ensure the page cannot be shorter than the viewport height

### LINKED ISSUE #1290: Homepage main container does not expand to fill page

thatbudakguy:
Fixes #1290


jcoyne:
This doesn't seem to be consistent with how we're doing it in other projects.  It seems like there's benefit to be gained from consistency. Should we put this in the component library or adjust the other projects?

thatbudakguy:
🤷🏻 I'm happy to put it in the component library. This is how I've always done it

PR #1368, opened 2024-09-04
fix footer

dnoneill:


jmartin-sul (approved):

jcoyne:
Unclear why this is necessary.  This overrides the expectations of the component-library: https://github.com/sul-dlss/component-library/blob/648047dda86e07ef6cea48affe225ddbd208d7c5/styles/footer.css#L11

dnoneill:
As discussed via Slack, this adds the classes that make it work with the component library.

PR #1374, opened 2024-09-04
no empty sidebar, codesnippets only when dataset

### LINKED ISSUE #1369: Code snippets should not appear on records for Scanned Maps or Collections
Code snippets should not appear on records for scanned maps or collections. 

In the case of scanned maps, the code is expecting a raster dataset but the object is a plain TIFF. 
https://earthworks-uat.stanford.edu/catalog/stanford-vs451mv3624

Collection objects have no data attached to them, so this code will fail if attempted.
https://earthworks-uat.stanford.edu/catalog/stanford-zr514ds1803

dnoneill:
closes #1369


hudajkhan:
@dnoneill Just a quick change required for removing the empty files. Thanks.

PR #1376, opened 2024-09-04
Update copy UI for citation modal

### LINKED ISSUE #1335: Citation modal and copy to match Code snippets modal and copy
Citation modal should match the Code snippets modal:

Close is the icon on the top right
Copy is on the top right (same as code snippets)
If the user clicks Copy, the button changes to a checkmark and Copied 

See here - https://www.figma.com/proto/nkgg905s5FyXoyPZjVMco0/EarthWorks?page-id=0%3A1&node-id=911-24473&node-type=FRAME&viewport=6758%2C4051%2C0.65&t=7tOM2cJm4ClC6AuQ-1&scaling=min-zoom&content-scaling=fixed&starting-point-node-id=911%3A24473&show-proto-sidebar=0

thatbudakguy:
Closes #1335 

This PR abstracts the controller that handled copying on the code snippets modal so that it can work for both the citation modal and code snippets modal, and then updates the UI for the citation modal to match.

Geoblacklight's own `clipboardController` doesn't include enough features to do this, but we can't redeclare a stimulus controller with that name, so the name was changed to `clipboardCopyController`. We could explore adding this functionality to Geoblacklight's version later.

Screenshot 2024-09-04 at 15 24 32


dnoneill:
Is there a reason for copying we aren't using the copy stimulus control from geoblacklight?

thatbudakguy:
Yes – what I meant above by
> Geoblacklight's own clipboardController doesn't include enough features to do this

Is that it doesn't support changing the target element to copy dynamically, which happens when you switch tabs on the code snippet modal.

It also expects an alert element to exist to update when you copy; we change the button text instead.

hudajkhan (approved):

PR #1408, opened 2024-09-12
Create hierarchical dates for opengeometadata records

### LINKED ISSUE #1407: Add hierarchical year indexing for non-Stanford content
Like #1136, but as a feature of `Earthworks::Harvester`, so that we can have hierarchical years for data coming from OpenGeometadata too.

thatbudakguy:
Closes #1407

Tested by using it to reindex OGM records in stage, where you can currently see the result:

Screenshot 2024-09-12 at 15 30 49

I made one or two very small changes from the traject implementation; notably other people aren't very good at providing their index years as integers...

### DIFF
--- a/lib/earthworks/harvester.rb
+++ b/lib/earthworks/harvester.rb
@@ -47,10 +47,16 @@ module Earthworks
         record.update({ 'schema_provider_s' => transformed_provider })
       end
 
+      # Filter out themes that are not in the controlled vocabulary from OGM
       if (themes = record['dcat_theme_sm'])
-        # Filter out themes that are not in the controlled vocabulary from OGM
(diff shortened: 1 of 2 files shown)


jcoyne:
Do we have any BCE dates we need to test?  Any non-Gregorian calendar dates? Would it make sense to limit the top end of the range to the max year in the corpus?

thatbudakguy:
I've just reindexed all of prod and all of the OGM data and there are no BCE dates. If there are non-gregorian dates, I guess we have no way of knowing, because none of them are 4-digit integers that are obviously outside the range of gregorian dates...but I think that'd be at least somewhat of an abuse of the spec, if not outright invalid.

Re: limiting the top end to the max year, that seems like a nice change, although I'm not sure offhand how to do it and it might make the facet logic a lot more complicated.

You can see what they look like unfurled in current production, since the reindex is done.

thatbudakguy:
No longer relevant after #1734.

PR #1409, opened 2024-09-12
Only add margin to consecutive paragraphs

thatbudakguy:
This overrides a geoblacklight style to prevent extra margin from being added to item descriptions.

# Before
Screenshot 2024-09-12 at 15 37 28

# After
Screenshot 2024-09-12 at 15 37 40

### DIFF
--- a/app/assets/stylesheets/earthworks.css
+++ b/app/assets/stylesheets/earthworks.css
@@ -365,3 +365,12 @@ body {
 #main-container {
   flex: 1;
 }
+
+/* Add margin for consecutive paragraphs in item descriptions */
+.document-metadata p {
+  margin-bottom: 0;
+}
+
+.document-metadata p + p {
+  margin-top: 1rem;
+}


jcoyne:
@dbranchini have you reviewed this?

@thatbudakguy which GeoBlacklight style are you referring to?  Should this be changed in GBL instead?

thatbudakguy:
Deployed to UAT.

jcoyne:
@thatbudakguy I'm confused what you're meaning when you wrote:
> This overrides a geoblacklight style 

but then 

> It's not a GBL override

Can you clarify that for me?

thatbudakguy:
Yes, sorry – it's confusing!

There are two declarations added in this PR that work together.

```css
/* Geoblacklight override: only add margin for consecutive paragraphs */
p {
  margin-bottom: 0;
}
```

This one is a Geoblacklight override, because Geoblacklight sets a global `p` style. However, it doesn't actually use `p` to render the description (it was unclear to me where, if at all, there are `p` in Geoblacklight, but I elected to leave it alone there).

```css
p + p {
  margin-top: 1rem;
}
```

This one is not a Geoblacklight override, because only Earthworks chooses to render descriptions as consecutive `p`.

jcoyne:
@thatbudakguy are you sure this isn't from Bootstrap? https://github.com/twbs/bootstrap/blob/fecd219983539aa0110a7e09d31b8aa4d5fb4348/scss/_reboot.scss#L130

jcoyne:
My concern here is that this is very broadly scoped. Can you do this only scoped to `.document-metadata p`?

thatbudakguy:
Sure, that makes sense. Updated.

jcoyne (approved):

PR #1410, opened 2024-09-12
Allow the main container to expand up to 1080p

thatbudakguy:
This small change benefits and effects only desktops, but allows
a little more breathing room for 1080p screens and above.


jcoyne:
@dbranchini have you reviewed this?

dbranchini:
Just to verify, this is allowing the header to stretch to the sides as it already does for main content area, correct @thatbudakguy ?

thatbudakguy:
Nope, the main content area doesn't currently do this. This is allowing the entire app to stretch up to a 1920px container (1080p size). The actual width will be slightly less, since that is inclusive of padding/margin/etc.

I noticed that the app often felt squished for no reason on my machine, and this seemed to help. If we don't want it, we can just close it.

PR #1478, opened 2025-01-27
Use a labeled <aside> for tools sidebar

### LINKED ISSUE #1437: Sidebar a11y updates: <aside>, aria-label
Because we have a sidebar on the left, the H1 is not seen/read first. We discussed this with SODA and they agreed that a left sidebar is a very common pattern (and in fact used on the WCAG site). Also, it appears under the category of "best practice" in SiteImprove, so it's not as important as an A or AA error. It also doesn't show up in Axe.

All that said, SODA suggested it would be improved by making the entire sidebar area an `<aside>` and giving it an aria-label that describes it, such as "Download, export and other options."

Screenshot 2024-10-16 at 5 06 42 PM

thatbudakguy:
This mirrors template changes in Blacklight 8.8 that convert the
sidebar from a <section> into an <aside> with label.

Closes #1437


jcoyne:
Can you link to the blacklight change?

thatbudakguy:
Updated to add it to the comment and commit message; good call

jcoyne (approved):

PR #1484, opened 2025-03-13
Update proxy to work with sul-embed

dnoneill:
This PR does a couple things:
1. fixes proxy response, the old way was returning a LocalJumpError https://app.honeybadger.io/projects/49895/faults/118259904. This is fixed by https://github.com/sul-dlss/earthworks/pull/1484/files#diff-4d95f4d583481eb175bbf04811a2ac3b579ea7c126965760942081923ed4da09R30 
2. Authorizes the user with a token from stacks instead of having to log in with shibboleth
3. Allows cross origin so sul-embed can request proxy

Part of https://github.com/sul-dlss/sul-embed/issues/2757

### DIFF
--- a/app/controllers/restricted_proxy_controller.rb
+++ b/app/controllers/restricted_proxy_controller.rb
@@ -26,8 +26,8 @@ class RestrictedProxyController < ApplicationController
       headers[k] = v
     end
 
-    self.status = proxied_response.status
-    send_data proxied_response.body, type: proxied_response.headers['Content-Type'], disposition: 'inline'
+    send_data proxied_response.body, type: proxied_response.headers['Content-Type'], disposition: 'inline',
+                                     status: proxied_response.status
   end
 
   private
--- a/app/controllers/restricted_proxy_controller.rb
+++ b/app/controllers/restricted_proxy_controller.rb
@@ -1,12 +1,33 @@
 class RestrictedProxyController < ApplicationController
-  before_action :authenticate_user!
+  before_action :headers_auth
+
+  def headers_auth
+    JWT.decode(request.params['stacks_token'], Settings.geo_proxy_secret, true, { algorithm: 'HS256' })
+    set_cors_headers
+  rescue JWT::ExpiredSignature
+    render json: { error: 'Expired token' }, status: :unauthorized
+  rescue JWT::DecodeError
+    render json: { error: 'Invalid token' }, status: :unauthorized
+  end
+
+  def set_cors_headers
+    origin = request.origin
+    permitted_origins = [Settings.cors.allow_origin_url]
+    if permitted_origins.include?(origin)
+      response.headers['Access-Control-Allow-Origin'] = origin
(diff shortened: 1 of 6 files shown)


thatbudakguy:
can we get some info about what this does and why, ideally as part of the commit message & copied here?

thatbudakguy:
I think what I want recorded somewhere is:
- the actual error that was occurring before (the specific name of it), and
- the change in this PR that fixes that error (adding `disposition: inline`? the CORS stuff? both?)

dnoneill:
@thatbudakguy done

thatbudakguy (approved):
perfect, thank you!

app/controllers/restricted_proxy_controller.rb
```diff
@@ -1,12 +1,33 @@
...
+  end
 
   def access
     proxied_headers.each do |k, v|
       headers[k] = v
     end
+
     self.status = proxied_response.status
-    self.response_body = proxied_response.body
+    send_data proxied_response.body, type: proxied_response.headers['Content-Type'], disposition: 'inline'
```
jcoyne, line 30: `send_data` also takes a `status:` argument.  Can we combine line 29 into here using it?

Also, isn't `proxied_response` an instance of `HTTP::Response` https://www.rubydoc.info/github/httprb/http/HTTP/Response Does that have a `headers` method?  I don't see it in their docs.

dnoneill: What headers are you talking about? The headers in set_cors_headers or the headers in    ``` proxied_headers.each do |k, v|
      headers[k] = v
    end```
    
 The first is for the controller. The second is what was previously in the code and I did not change it.

jcoyne: I'm specifically speaking about `proxied_response.headers['Content-Type']`

spec/controllers/restricted_proxy_controller_spec.rb
```diff
@@ -12,29 +12,30 @@
 
   context 'with a current user' do
     before do
-      sign_in create(:user)
       allow(HTTP).to receive(:get).and_return(
         double(:response, {
```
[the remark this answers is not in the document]
thatbudakguy:
```suggestion
        instance_double(HTTP::Response, {
```
like this?

app/controllers/restricted_proxy_controller.rb
dnoneill: Also I double checked  proxied_response.headers and it returns #<HTTP::Headers:0x0000000154359738> so that is a valid field.

jcoyne: Cool. Thanks for checking. Switching to an instance double 👇 would prevent doubts like this.

dnoneill: I have done that.

jcoyne (approved):

PR #1507, opened 2025-06-11
add resource type and spatial coverage to json api

dnoneill:
closes https://github.com/sul-dlss/sul-bento-app/issues/775


app/controllers/catalog_controller.rb
```diff
@@ -13,6 +13,14 @@ class CatalogController < ApplicationController
     BotChallengePage::BotChallengePageController.bot_challenge_enforce_filter(controller, immediate: true)
   end
 
+  before_action only: :index do
```
thatbudakguy, line 16: Can you put a comment on this saying that Bento calls the JSON API and needs the icon info?

PR #1509, opened 2025-06-13
Allow JSON requests to go unchallenged by turnstile

jcoyne:


jcoyne:
Just do bento dev while on the VPN if you want to connect to earthworks.  The safelist allows VPN ips through.

PR #1510, opened 2025-06-13
add test for header icons component

dnoneill:


spec/components/header_icons_component_spec.rb
```diff
@@ -0,0 +1,52 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+require 'spec_helper'
```
jcoyne, line 4:
```suggestion
```

rails_helper already includes spec helper.

jcoyne (changes requested):

PR #1533, opened 2025-08-18
Upgrade to Rails 8

### LINKED ISSUE #1532: Upgrade to Rails 8

thatbudakguy:
Closes #1532


public/icon.svg
```diff
@@ -0,0 +1,3 @@
+<svg width="512" height="512" xmlns="http://www.w3.org/2000/svg">
```
jcoyne, line 1: I don't think we want either of these icons.

public/robots.txt
```diff
@@ -1,5 +1 @@
-# See http://www.robotstxt.org/robotstxt.html for documentation on how to use the robots.txt file
-#
-# To ban all spiders from the entire site uncomment the next two lines:
-User-agent: *
-Disallow: /
```
jcoyne, line 5: Was this change intentional?

thatbudakguy: Nope! Good catch. Reverted it.

jcoyne (approved):

PR #1539, opened 2025-08-21
Rename the background job role to 'cron'

thatbudakguy:


config/deploy.rb
```diff
@@ -38,7 +38,8 @@
 
 set :honeybadger_env, fetch(:stage)
 
-set :whenever_roles, [:whenevs]
+# Run scheduled jobs on the background workers
+set :whenever_roles, [:background]
```
jcoyne, line 42: Whenever is not the same as background jobs.  I think a better name is "cron".  We probably don't want crons on more than one machine, but background jobs you often do want on multiple machines.

thatbudakguy: Updated the name to `cron`.

jcoyne (approved):

PR #1541, opened 2025-08-25
Update rubocop cops and autocorrect

thatbudakguy:
Rubocop had been complaining about all of the new cops, so this
activates them and fixes the suggested changes.

### DIFF
--- a/spec/features/also_available_spec.rb
+++ b/spec/features/also_available_spec.rb
@@ -11,6 +11,7 @@ describe 'Also available at urls' do
   context 'when a document does not have a searchworks url' do
     it 'does not shows it' do
       visit solr_document_path 'stanford-cg357zz0321'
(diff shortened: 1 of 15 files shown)


spec/features/also_available_spec.rb
```diff
@@ -11,6 +11,7 @@
   context 'when a document does not have a searchworks url' do
     it 'does not shows it' do
       visit solr_document_path 'stanford-cg357zz0321'
+      expect(page).to have_content '10 Meter Contours: Russian River Basin, California'
       expect(page).to have_no_link 'SearchWorks'
```
thatbudakguy, line 15: These changes are because of a new cop that doesn't like capybara negation matchers being used immediately after page load (presumably because they can give a false positive). I added a trivial matcher to check that the page title is correct before attempting a negation matcher.

PR #1546, opened 2025-09-10
Pull in Geoblacklight v5.1 styles

thatbudakguy:
- **Update the styles and JS to GBL v5.1**
- **Add a non-Stanford paper map fixture for testing**

### DIFF
--- a/app/helpers/earthworks_geoblacklight_helper.rb
+++ b/app/helpers/earthworks_geoblacklight_helper.rb
@@ -18,9 +18,4 @@ module EarthworksGeoblacklightHelper
   rescue License::UnknownLicenseError
     tag.span args[:value]
   end
-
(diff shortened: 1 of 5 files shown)


app/helpers/earthworks_geoblacklight_helper.rb
```diff
@@ -18,9 +18,4 @@ def render_license(args)
   rescue License::UnknownLicenseError
     tag.span args[:value]
   end
-
-  # Use Mirador as the IIIF manifest viewer
-  def iiif_manifest_viewer
-    tag.div(nil, id: 'mirador', data: { 'manifest-url' => @document.viewer_endpoint })
-  end
```
thatbudakguy, line 25: Override no longer needed; it became the default in GBL 5.1

PR #1558, opened 2025-10-30
Just show the description for search results

thatbudakguy:
Because we added several other fields into what's normally shown when
detailed results are enabled on the search results page, you could get
strange-looking entries with tons of years or publisher info.

This just keeps it to the actual description.


thatbudakguy:
# before
Screenshot 2025-10-29 at 17 23 17

# after 
Screenshot 2025-10-29 at 17 24 22

hudajkhan (approved):

PR #1559, opened 2025-10-30
Sort by recent items when there is no active query

thatbudakguy:
Our 'relevance' sort uses the search term (if there is one), then
the title. Because of this, when there is no query, you just get things
that are alphabetically first, which is mostly items in other languages
due to unicode order of precedence, like our Gaihozu index maps.

This way, the first thing you see if you don't put in a query is the
items that have more recent dates, which are more immediately useful.


thatbudakguy:
# before 
Screenshot 2025-10-29 at 17 29 36

# after
Screenshot 2025-10-29 at 17 29 56

PR #1578, opened 2026-01-06
Use an appropriate solr schema version.

cbeer:
I don't know where `4.0` came from, but `1.7` is current (with some breaking changes).


config/solr_configs/schema.xml
```diff
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<schema name="geoblacklight-schema" version="4.0">
+<schema name="geoblacklight-schema" version="1.6">
```
dnoneill: This is coming from https://github.com/geoblacklight/geoblacklight/blob/7d929e3ef570c899ceb4653243d357d289d8507e/solr/conf/schema.xml#L2. We should probably fix it there too.

PR #1624, opened 2026-05-06
Rearrange record show page element order

thatbudakguy:
This reorganizes the record detail page to look similar to Searchworks.
The record action sidebar is on the right, and the record details take
up a larger space on the left, including the map preview.


thatbudakguy:
Screenshot 2026-05-06 at 16 15 18

jcoyne (approved):

Issue #1531, opened 2025-08-18
Shift SDR indexing process to Earthworks infrastructure


thatbudakguy:
This is a top-level tracking ticket for work outlined in [ADR #0024](https://github.com/sul-dlss-labs/engineering-design-adrs/blob/main/0024-earthworks-indexing.md), "Transfer Earthworks indexing process to Earthworks codebase and VMs".

Misc. tasks:

- [ ] Test releasing records from argo-stage and indexing into earthworks-stage
- [ ] Update the DevOpsDocs

PR #1634, opened 2026-05-28
Update cocina_display

cbeer:
It's not clear to me whether this code is actually used right now?


thatbudakguy:
It isn't – it was prep for https://github.com/sul-dlss/earthworks/issues/1531 (specifically, https://github.com/sul-dlss/earthworks/issues/1535) but I never got that far.

There's a chance it gets revisited in the upcoming workcycle, but we can also take it out until/unless that happens.

thatbudakguy (approved):

PR #1639, opened 2026-06-02
Index from SDR

jcoyne:


app/services/cocina_service.rb
```diff
@@ -1,8 +1,14 @@
 # Retrieve public Cocina records from PURL
 class CocinaService
   # Fetch the public Cocina JSON and parse into a hash
-  # TODO: https://github.com/sul-dlss/earthworks/issues/1537
   def self.fetch_record(druid)
     Rails.logger.info "Fetching metadata for druid:#{druid}"
+    response = HTTP.get("https://#{Settings.purl.hostname}/#{druid}.json")
+    return nil unless response.status.success?
```
thatbudakguy: Should we make an unsuccessful response also raise, so it gets logged below?

app/services/solr_service.rb
```diff
@@ -1,14 +1,17 @@
...
+    Blacklight.default_index.connection.commit
   end
 
   # Update a record in the index
-  # TODO: https://github.com/sul-dlss/earthworks/issues/1536
   def self.update(record)
-    Rails.logger.info "Updating Solr document: #{record.druid}"
+    Rails.logger.info "Updating Solr document: stanford-#{record.druid}"
+    doc = CocinaToSolrMapper.map(record)
+    Indexer.new.solr_update([doc])
```
thatbudakguy: I should've specified this in the ticket, but I think we should just ditch the existing `Indexer` and move its functionality here (i.e. this method should also call `Blacklight.default_index`, and the `commitWithin` config should move here).

I think it was only ever used to update the index as part of GeoMonitor, and that's turned off (and the associated code has been removed).

app/services/cocina_to_solr_mapper.rb
```diff
@@ -0,0 +1,180 @@
...
+            end
+          elsif k == value.to_s
+            match = v
+            break
+          end
+        end
+        match
+      end.uniq
+    end
+  end
```
thatbudakguy: I like having this available for other indexing configs to use too! Could we get it in a separate file and maybe with some minimal tests?

app/services/cocina_service.rb
jcoyne: Done.

app/services/solr_service.rb
jcoyne: It's still used in the geomonitor:reset_availability rake task.

thatbudakguy: Geomonitor is turned off because it's not compatible with Aardvark records, and we've already removed other parts of code needed to make it work. Let's remove that rake task, since it shouldn't be invoked anyway.

spec/services/translation_map_spec.rb
thatbudakguy, line 1: thank you, this is great!

app/services/solr_service.rb
thatbudakguy: (And if it's installed as a gem, we should remove it from the Gemfile).

thatbudakguy (approved):

PR #1642, opened 2026-06-02
Fix solr configuration for sort fields

### LINKED ISSUE #1641: dct_title_sort is not sortable
org.apache.solr.common.SolrException: can not sort on a field w/o docValues unless it is indexed=true uninvertible=true and the type supports Uninversion: dct_title_sort

jcoyne:
Fixes #1641


thatbudakguy (approved):

thatbudakguy:
Does this also need to be reflected in sul-solr-configs?

PR #1675, opened 2026-06-22
Combine all sections of the record view sidebar

### LINKED ISSUE #1629: Combine all sections in the record view sidebar
This will match Searchworks's sidebar, which doesn't have two distinct sections. It will also solve a longstanding bug where the relations panel doesn't load in, or doesn't load fast enough, and there is a "ghost" empty second section displayed.

If we combine the sections, we can get rid of the sidebar component that houses all of them, and just render all of the child components in a list in the parent component instead of the sidebar component.

jcoyne:
Fixes #1629

### DIFF
--- a/app/components/document/sidebar_component.html.erb
+++ b/app/components/document/sidebar_component.html.erb
@@ -1,3 +1,5 @@
 <div class="bg-fog-light component-container rounded">
-  <%= safe_join filtered_components %>
+  <% components.each do |component| %>
+    <%= render component, document: %>
+  <% end %>
 </div>
--- a/app/components/document/sidebar_component.html.erb
+++ b/app/components/document/sidebar_component.html.erb
@@ -1,13 +1,3 @@
-<% filtered_components = filter_components(components) %>
-<% if filtered_components.any? %>
-  <div class="bg-fog-light component-container rounded">
-    <%= safe_join filtered_components, separator %>
-  </div>
-<% end %>
-
-<% filtered_components = filter_components(secondary_components) %>
-<% if filtered_components.any? %>
-  <div class="bg-fog-light component-container rounded mt-4">
-    <%= safe_join filtered_components, separator %>
-  </div>
-<% end %>
+<div class="bg-fog-light component-container rounded">
+  <%= safe_join filtered_components, separator %>
+</div>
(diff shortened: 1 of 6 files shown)


app/components/document/sidebar_component.rb
```diff
@@ -9,22 +9,20 @@ def components
...
 
     # Determine if there's actually anything to show by rendering everything
-    def filter_components(components)
-      components.map { |component| render component, document: }.compact_blank
+    def filtered_components
+      @filtered_components ||= components.map { |component| render component, document: }.compact_blank
+    end
+
+    def render?
+      filtered_components.any?
```
thatbudakguy: I think both of these can actually just go away, since the addition of the ShowToolsComponent means there is always at least one thing to render, and so we don't even need to check `render?`.

thatbudakguy (commented):
Is it worth leveraging the ViewComponent solution for inserting the separator, and then getting rid of `filtered_components` and the `safe_join`?

I've never tried this, but I think it would go like:
- Make a new component called like `SidebarPanelComponent`; all it is is the `bg-fog-light component-container rounded` wrapper div that renders its contents inside. If its contained component doesn't render, it doesn't render
- Make a `SidebarSeparatorComponent` that's just the `<hr>` with classes
- The `SidebarComponent` template becomes:

```erb
<%= render(SidebarPanelComponent.with_collection(components, spacer_component: SidebarSeparatorComponent.new)) %>
```

jcoyne:
If it were up to me, I'd do away with `hr` and just add the line via css

jcoyne:
I don't understand your proposed method. Typically with_collection is passed a collection of model objects, not of components.

thatbudakguy:
> If it were up to me, I'd do away with `hr` and just add the line via css

I like this even better, tbh.

app/components/document/sidebar_component.html.erb
```diff
@@ -1,3 +1,3 @@
 <div class="bg-fog-light component-container rounded">
-  <%= safe_join filtered_components, separator %>
+  <%= safe_join filtered_components %>
```
thatbudakguy: Can we just do:
```erb
<% components.each do |component| %>
  <%= render component %>
<% end %>
```
Instead of prerendering them with `filtered_components`?

jcoyne: sure

thatbudakguy (approved):
Thanks for being willing to iterate on this!

PR #1680, opened 2026-06-23
Remove unnecessary method override

jcoyne:


thatbudakguy:
I think we're overriding this because it's used a couple places in Geoblacklight: https://github.com/search?q=repo%3Ageoblacklight/geoblacklight%20document_available%3F&type=code

jcoyne:
@thatbudakguy it's not necessary, because the GBL method cover the same cases as this one, right?

https://github.com/geoblacklight/geoblacklight/blob/284aff55a98331d13d46f7f940cc521f817e21dd/app/helpers/geoblacklight_helper.rb#L4-L6
https://github.com/geoblacklight/geoblacklight/blob/284aff55a98331d13d46f7f940cc521f817e21dd/app/models/concerns/geoblacklight/solr_document.rb#L35-L37

thatbudakguy:
Yeah honestly I don't super understand what's going on here since we aren't providing a separate method for `document.available?` which is the only difference between the two implementations. So I guess we can just remove this.

thatbudakguy (approved):

PR #1693, opened 2026-06-25
Remove obsolete tasks

thatbudakguy:
- **Remove bin/build-css**
  We no longer build GeoBlacklight's stylesheet using SCSS.
  

- **Remove package.json and package-lock.json**
  All of our JS dependencies are delivered via importmap, and all
  of our CSS dependencies are delivered via gems/CDN.
  

- **Remove earthworks:server Rake task**
  It's identical to the geoblacklight:server task that is already
  provided by the gem.
  

- **Remove earthworks:install Rake task**
  All it does is run db:migrate and it isn't called by anything.


jcoyne (changes requested):
I believe this change breaks `bin/dev`

https://github.com/sul-dlss/earthworks/blob/main/bin/dev#L11

thatbudakguy:
I thought the only reason we had that was to run Foreman, but there's only one task for Foreman to run, so it seemed unnecessary?

I guess we could just restore the Procfile if you want to run the Rails server that way 🤷🏻

thatbudakguy:
Restored Procfile.dev.

jcoyne (approved):

PR #1712, opened 2026-07-08
Update GeoBlacklight to v6.0.0-alpha.4

thatbudakguy:
[edited]


Gemfile
```diff
@@ -61,8 +61,7 @@ group :production do
 end
 
 gem 'blacklight', '~> 9.0'
-gem 'rsolr' # required for Blacklight
```
jcoyne, line 64: I don't think we want to remove this. The generator in Blacklight adds it for you.  geoblacklight should not bundle it. Furthermore it also adds it here: https://github.com/geoblacklight/geoblacklight/blob/551921986fa05031f880a8c44730ed507aa57e92/lib/generators/geoblacklight/install_generator.rb#L65-L67

jcoyne (changes requested):

jcoyne (approved):

thatbudakguy:
I think https://github.com/geoblacklight/geoblacklight/pull/1861 may fix the aXe error.

PR #1716, opened 2026-07-15
Remove okcomputer check for download cache

thatbudakguy:
Generated downloads were removed from GBL core.


thatbudakguy:
Handled by #1715

PR #1720, opened 2026-07-16
Display file format as simple text in downloads

thatbudakguy:
The badges don't play nice with our sidebar styling, and end up
looking weird when the file name is long.


thatbudakguy:
Screenshot 2026-07-16 at 13 32 15

PR #1723, opened 2026-07-16
Use the same layout & width as DataWorks

thatbudakguy:
In https://github.com/sul-dlss/earthworks/commit/9c90b81b82adb65abd58f1974e23664858a58ab3, we changed the layout to give more breathing room to the map and sidebar.

Now that we've rearranged the UI to match Searchworks and Dataworks, and with some of the changes in GBL6, this is no longer necessary – there's plenty of space for the viewer and metadata to breathe.

This reverts https://github.com/sul-dlss/earthworks/commit/9c90b81b82adb65abd58f1974e23664858a58ab3 and makes the arrangement of the columns on the show page match our other platforms.

### DIFF
--- a/app/assets/stylesheets/earthworks.css
+++ b/app/assets/stylesheets/earthworks.css
@@ -4,6 +4,20 @@
   --stanford-palo-alto-rgb: 23, 94, 84;
 }
 
+/* Below lg, match the gap between show-document and the sidebar to the metadata section gap. */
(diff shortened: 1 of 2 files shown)


thatbudakguy:
# Before
Screenshot 2026-07-16 at 15 10 45

# After
Screenshot 2026-07-16 at 15 10 35

jcoyne (approved):

PR #1739, opened 2026-07-23
Update the tests for the Cocina mapper

thatbudakguy:
This replaces the tests with real objects so we can make better
assertions about the important fields like resource class and
type, as well as real file paths getting generated in references.

Also makes a few tweaks to the mapper, like fixing class/type for
index maps and adding a FlatGeoBuf download for objects that
have it.


app/services/cocina_to_solr_mapper.rb
```diff
@@ -80,7 +82,10 @@ def map_resource_class
   end
 
   def map_resource_type
-    TranslationMap.new('geo_resource_type').translate(record.all_forms.map(&:to_s) + record.subject_topics)
+    res_types = TranslationMap.new('geo_resource_type').translate(record.all_forms.map(&:to_s) + record.subject_topics)
+    return res_types.without('Polygon data', 'Point data') if index_map? # Special case: don't treat it as a dataset
```
thatbudakguy, line 86: Index maps are always one of these two types, but we don't want to categorize them the same way as actual polygon or point data; the "data" is just pointers to scanned maps or other objects

app/services/cocina_to_solr_mapper.rb
```diff
@@ -80,7 +82,10 @@ def map_resource_class
   end
 
   def map_resource_type
-    TranslationMap.new('geo_resource_type').translate(record.all_forms.map(&:to_s) + record.subject_topics)
+    res_types = TranslationMap.new('geo_resource_type').translate(record.all_forms.map(&:to_s) + record.subject_topics)
```
jcoyne, line 85: Can we filter the forms just to the OGM ones (or exclude the LOC forms) and only translate those? It seems confusing to have a x->x translation for LOC.

thatbudakguy: wouldn't that just be splitting the config into two separate files and two method calls? we still need to keep the list of LOC terms somewhere.

this same approach was used for [theme](https://github.com/sul-dlss/searchworks_traject_indexer/blob/c2a53af3660a0fee124adb7a64da640f653e8889/lib/translation_maps/geo_theme.yaml#L6-L26) and [format](https://github.com/sul-dlss/searchworks_traject_indexer/blob/main/lib/translation_maps/geo_format.yaml) too, because in those cases we also have some values that need translating and some that don't. It looks like we still need to port those configs to Earthworks, which doesn't have the full versions.

jcoyne: Why do we need to keep a list of LOC terms too?  Can we just use the term as given if they are in the LOC scheme?

thatbudakguy: Ah, I didn't understand before - you mean filtering the Cocina using other attributes to check if it's already LOC? That might be possible, provided the list of terms is 1:1

jcoyne: Yes, precisely.  We store which schema every term is from, so it seems doable.

thatbudakguy: Just checked and unfortunately this isn't as easy as it seems. The values allowed in Aardvark are LoC "Cartographic Genres", which are just a small subset of LCSH.

For example, https://purl.stanford.edu/qc464mj0004 has the subject "Statistical maps", which is valid per Aardvark. But it also has subjects like "Recreational" and "Historical geography" which are _not_ valid. All of them are of type `topic` and have source `code` set to `lcsh`, so there's no way to distinguish.

I think this means we have to enumerate all of the allowed values, like we do now.

PR #1743, opened 2026-07-30
Restore the translation mapping for format and theme

thatbudakguy:
This ports config that was remaining in searchworks-traject-indexer,
so that we can delete it from there.


thatbudakguy:
Also updates `TranslationMap` to handle multiple output values, because the old version did.

jcoyne (approved):

PR #1745, opened 2026-07-30
Update GeoBlacklight to v6.0.0-alpha.7

thatbudakguy:


app/assets/stylesheets/earthworks.css
```diff
@@ -18,19 +18,6 @@
   }
 }
 
-.header-icons {
```
thatbudakguy, line 15: this class is no longer used in GBL.

app/assets/stylesheets/earthworks.css
```diff
@@ -212,29 +197,10 @@
...
-
 .search-query-form .search-btn .blacklight-icons {
   width: auto;
 }
 
-/* .search-query-form .search-btn .blacklight-icons-search {
-  display: none;
-} */
-
-#main-container #leaflet-viewer {
```
thatbudakguy, line 216: these styles moved upstream into GBL.

app/components/header_icons_component.rb
thatbudakguy, line 1: we don't need to override this any more.

app/helpers/earthworks_blacklight_helper.rb
thatbudakguy, line 1: overrides here were only for setting width classes on the body; no need to override any longer.

jcoyne (approved):

PR #1758, opened 2026-08-12
Add full reindexing script

thatbudakguy:
This adds a helper script for doing a full reindex in local dev
(or prod, if needed).

It borrows the approach from
https://github.com/sul-dlss/searchworks_traject_indexer/pull/1511,
which means that PR can be closed.


jcoyne:
Is there a reason you wouldn't just reset the kafka offsets?

thatbudakguy:
This is way, way faster and avoids replaying a massive number of records that we don't care about.

It's also mostly for dev use, where you don't want to be replaying just to update your local solr's records.

jcoyne:
I think splitting the purl-fetcher index into a specific topic for earthworks will solve that.

thatbudakguy:
What do you mean? Resetting offsets affects every consumer – it's how you replay things in production. It isn't the right mechanism for local development – you would need to pause or kick out the prod consumer before attaching your own consumer to make it work. That's mostly what this PR is for.

jcoyne:
Each consumer has it's own group.  You can have a different consumer group for different environments.

script/reindex_sdr.rb
```diff
@@ -0,0 +1,120 @@
...
+
+# Fake a Kafka message saying to index this druid
+Message = Struct.new(:key, :value)
+def create_message(druid:)
+  key = druid
+  value = { true_targets: ['Earthworks'] }.to_json
+  Message.new(key, value)
+end
+
+reader = PurlFetcher::Client::Reader.new
```
jcoyne, line 101: Please don't connect to purl-fetcher like this. This adds extra load on the purl-fetcher. We can get this data from the kafka topic.

PR #1769, opened 2026-09-10
Treat scanned maps as Maps instead of Raster data

thatbudakguy:
These were created as georeferenced derivatives of scanned maps;
despite going through the GIS pipeline and having the 'geo'
content type, they are maps, not raster datasets.


jcoyne:
This seems like it ought to be two different PRs, right?

thatbudakguy:
Fixed; I had rebased so that I could borrow #1758 to test out these config changes in development.

jcoyne (approved):

PR #1775, opened 2026-09-16
Synchronize the hide/show descriptions button with expand/collapse

thatbudakguy:
GBLv6 has a native truncation solution for descriptions that leaves
the full description in the page. This means that our existing
hide/show descriptions button ended up showing the full description
for every item.

This combines the two methods: when descriptions are hidden, the UI
looks as it did before. When they are shown, each record gets a
truncated description, which can then get individually expanded or
collapsed if the user wants to read more.

Also fixes an a11y issue where the hide/show descriptions button
was not keyboard-selectable and had no focus ring.


thatbudakguy:
When "show descriptions" is on, we still truncate so that you can expand/collapse each one:

Screenshot 2026-09-16 at 15 38 38

app/assets/stylesheets/earthworks.css
```diff
@@ -189,7 +178,6 @@
 }
 
 .navbar-logo {
-  /* TODO: remove. Should be fixed when geoblacklight uses blacklight v9 */
   background-size: auto;
```
jcoyne, line 181: can we remove this?

app/assets/stylesheets/earthworks.css
```diff
@@ -223,26 +211,33 @@
...
-.expand-collapse .btn-outline-primary {
-  display: flex;
-  &.active {
+.btn-check:focus-visible + .btn {
+  outline: 2px solid var(--bs-btn-focus-box-shadow);
+  box-shadow: 0 0 0 4px var(--stanford-black);
+}
+
+.toggle-descriptions {
+  & .btn-outline-primary {
```
jcoyne: why are there leading &?

app/assets/stylesheets/earthworks.css
```diff
@@ -291,8 +286,8 @@
   border-radius: 0.25rem;
 }
 
-.document-counter {
-  display: none;
+.document-main-section .metadata {
+  padding: 0;
```
jcoyne, line 290: Why not use the .p-0 bootstrap class on that element?

jcoyne (changes requested):

app/assets/stylesheets/earthworks.css
thatbudakguy: I tried it and it made the logo the wrong size, so it apparently is still required in v9. I removed the comment instead.

app/assets/stylesheets/earthworks.css
thatbudakguy: because `!important` is evil

app/assets/stylesheets/earthworks.css
thatbudakguy: dunno! removed them.

jcoyne (approved):

